PR #139, opened 2024-01-29
Minor changes in the header

ChahidLaaouar:
The Log in screen close modal button text is now a x instead of 'Close Modal'. The close modal button no longer rotates when hovered.


ChahidLaaouar:
Logo in header redirects to the home page when clicked instead of being unclickable

PR #1979, opened 2025-05-26
Removes temporary branch from pipelines.

Sandervg03:
Removes temporary branch from pipelines. Also adds conditional to wether you want to generate an API Gateway.


Ryan-Reddy (approved):
grmbl vanwaar de codestyle differences?

PR #3022, opened 2026-05-20
fix: make sidebar sticky on host and guest dashboard

YusufCetinkaya1135:
Issue #3021 – Fix sticky sidebar on guest and host dashboard
The sidebar in both the guest and host dashboards was not sticky. When scrolling down the page, the sidebar would scroll along with the content instead of staying in place.
This fix makes the sidebar remain in a fixed position during scroll on both dashboards.


OfficialLachkid (approved):
LGTM

PR #3164, opened 2026-09-01
perf(ci): turn on the npm cache and stop paying for superseded runs

roberthopman:
## Close or Relate the Issue
- Closes #
- Related Issue: # (stacked on #3163, which must merge first)

## Proposed Changes
Description:

**Stacked on #3163.** Until that merges this PR's diff also shows its commit; afterwards it shrinks to the two commits below.

Install is the single biggest cost in the frontend pipeline. From the last green run on #3163:

```
Step                                    Time    Share
----                                    ----    -----
actions/checkout@v4                       3s
Setup Node.js environment                 1s      ◀ no cache restored
Install dependencies                    160s      48%
Run Tests                                41s      12%
Build release                           122s      37%
Post Setup Node.js environment            0s      ◀ nothing saved
                                        ----
                                        333s
```

**The finding.** Both frontend workflows already passed `cache-dependency-path` to `setup-node`, but neither passed a `cache:` input. `cache-dependency-path` only computes the cache key; without `cache: npm` the action caches nothing. Those two `0s` lines are the restore and save steps doing no work. All 2801 packages have been downloaded on every run since the config was added.

**Two commits, refactor separated from behaviour** per the `## Commits` rule in `CLAUDE.md`.

`caa3c7e` extracts the frontend setup into `.github/actions/setup-frontend`. `node-version` was declared in four workflows and the npm pin in two, kept in sync by hand. That had already gone wrong: the missing `cache:` was copy-pasted from `webapp-build-test.yml` into `ci.yml` along with everything else. No behaviour change, except `ci.yml`'s install drops `--verbose`, which only made the log longer. `cache: npm` now appears exactly once in the repository.

`e092620` enables the cache, adds `--prefer-offline --no-audit --fund=false`, and adds a `concurrency` block.

**Concurrency cancels on `pull_request` only.** For a PR `github.ref` is `refs/pull/N/merge` so runs group per PR, which is where the waste is. For a push, all pushes to `acceptance` share one group, so `cancel-in-progress: true` would let a later merge kill an earlier merge's build. `CLAUDE.md` says to revert a merge whose pipeline fails, and you cannot revert on a signal you cancelled.

**Expected effect.** Caching `~/.npm` removes download time but not extract and link, so install should fall to roughly 70-100s rather than near zero. That is an estimate; the run on this PR is the measurement. Caching `node_modules` outright would cut it to seconds, but a stale entry produces confusing failures and it needs a hand-maintained key. Worth revisiting once there is a real number.

**Not addressed.** `Build release` is 122s of `react-scripts build`, the next biggest number. The real fix is migrating to Vite, which the repo already uses for Cypress. That is a project, not a CI tweak.

`--no-audit` removes the audit call from CI. It was not a gating step, Dependabot covers the same ground, and the PR template treats `npm audit` as a manual checkbox.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
`caa3c7e` moves existing steps into a composite action without changing what they do. Read it as a no-behaviour-change diff and review only `e092620` for behaviour.
- .github/workflows/ci.yml
- .github/workflows/webapp-build-test.yml

## Evidence

No UI change. Evidence is the step timing table above, taken from run 33505375984. The `build` check on this PR is the after measurement; compare its `Install dependencies` against the 160s baseline.

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

No UI change.

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

No dependency changes. `package.json` and `package-lock.json` are untouched.

## Checklist
- [x] Conventions
- [ ] Testing
  - [x] Code tested locally (YAML validated; timings measured from run 33505375984)
  - [ ] Jest tests are included
  - [ ] Jest tests are passing

A CI change is not unit-testable. It is verified by this PR's own checks.

## Keep or delete my branch
- [ ] Keep my branch

### DIFF
--- a/.github/actions/setup-frontend/action.yml
+++ b/.github/actions/setup-frontend/action.yml
@@ -1,0 +1,22 @@
+name: Setup frontend
+description: >
+  Node, npm and dependencies for frontend/web. Declared once so the version
+  and the install flags cannot drift between workflows.
+
+runs:
+  using: composite
+  steps:
+    - name: Setup Node.js environment
+      uses: actions/setup-node@v6.0.0
+      with:
+        node-version: 22
+        cache-dependency-path: frontend/web/package-lock.json
+
+    - name: Pin npm to the version Dependabot generates lockfiles with
+      shell: bash
+      run: npm install -g npm@11.6.2 --ignore-scripts
+
+    - name: Install dependencies
+      shell: bash
+      working-directory: frontend/web
+      run: npm ci
--- a/.github/actions/setup-frontend/action.yml
+++ b/.github/actions/setup-frontend/action.yml
@@ -10,6 +10,7 @@ runs:
       uses: actions/setup-node@v6.0.0
       with:
         node-version: 22
+        cache: npm
         cache-dependency-path: frontend/web/package-lock.json
 
     - name: Pin npm to the version Dependabot generates lockfiles with
@@ -19,4 +20,4 @@ runs:
     - name: Install dependencies
       shell: bash
       working-directory: frontend/web
-      run: npm ci
+      run: npm ci --prefer-offline --no-audit --fund=false
--- a/.github/actions/setup-frontend/action.yml
+++ b/.github/actions/setup-frontend/action.yml
@@ -20,4 +20,4 @@ runs:
     - name: Install dependencies
       shell: bash
       working-directory: frontend/web
-      run: npm ci --prefer-offline --no-audit --fund=false
+      run: npm ci --prefer-offline --no-audit --fund=false --ignore-scripts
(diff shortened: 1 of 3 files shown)


BeMounir (commented):
smart fix. github actions was reinstalling packages for no reason so i get the change. i do not really understand the concurrency part very well but i think thats just me. sonarcloud is giving another warning like the one in the other #3163 where you need to add --ignore-scripts for safety. reducing the load times by 70-100 seconds is substantial. i would listen to sonarcloud and make the small tweak though

BeMounir (approved):

PR #3168, opened 2026-09-02
#3038 Add server-side quote engine for direct booking websites

karimh01:
## Close or Relate the Issue
[//]: # (Only if this should close the issue)
- Closes #

[//]: # (in the issue a reference will be added to this pr)
- Related Issue: #3038

## Proposed Changes
[//]: # (Add a detailed description of the changes here)
Description:

Adds the server-side quote engine for direct booking websites. No endpoint, no gateway wiring, no frontend. This is the business layer only, so nothing calls it yet.

What it does:
* Validates the date range: real calendar dates, checkout after checkin, not in the past, and a 366 night engine cap
* Validates guest count against property capacity
* Applies stay restrictions (MinimumStay, MaximumStay, MinimumAdvanceReservation, MaximumAdvanceReservation)
* Checks availability the same way the bookings Lambda does, so a quote cannot say yes to dates the booking guard would reject
* Calculates the price breakdown in cents
* Signs the quote so the price cannot be changed in the browser, and expires it after 30 minutes so an old price cannot be reused

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
[//]: # (Indicate the size of this change.)
[//]: # (Clarify under Refactoring which parts are moved/unchanged code, so the reviewer doesn’t review old logic.)
- [ ] Huge change (1000+ lines, mostly refactored/moved code. Clarify in [Refactoring](#Refactoring))
- [x] Big change (+-max 1000)
- [ ] Small change (less than 300)

## Refactoring
Refactors following files, but didn't change the code. This is to avoid reviewing old logic.
N/A, all new files

## Evidence
- [ ] Screenshots or screen recording added for UI changes
- [ ] Before/after images added when relevant
- [ ] Images clearly show the implemented change
- [ ] Image quality is readable (no cropped or blurry evidence)

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop
- [ ] No broken layout, overflow, or hidden content on tested screen sizes

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated
- [ ] Checked for vulnerabilities using "npm audit"

## Checklist
[//]: # (All boxes must be checked before merging.)
- [x] Pull request has at least 2 reviewers assigned
- [x] PR title is descriptive and includes issue number
- [x] Conventions
  - [ ] No config files have been added (Amplify config, cache files)
[...]

Issue #3038, opened 2026-05-28
Use Case / Scenario Testing Direct Booking Websites


stefanhopman1:
Use Case / Scenario Testing Direct Booking Websites enables Domits to provide a dedicated, branded website experience outside the main platform while still using the same core inventory, pricing, availability, and booking infrastructure.

# 🧑‍💼 Roles

- Guest / Traveler - Luxury traveler booking directly through the website.
- Host / Property Owner - Owner managing one or multiple listings.
- Property Manager - Professional operator managing portfolios.
- Admin / Website Operator - Manages platform settings, pages, branding, and operational workflows.
- Marketing Manager - Manages SEO, campaigns, lead capture, and analytics.
- Concierge / Guest Experience Team - Supports guests before, during, and after stays.
- Maintenance / Housekeeping Staff - Handles operational property workflows.
- Finance / Accounting Manager - Manages transactions, invoices, payouts, taxes, and reporting.

---

# 🌐 1. Website & Branding

## Host Stories

- [x] As a host, I want to customize my website branding so it matches my hospitality business identity.
- [x] As a host, I want to upload my logo, colors, fonts, and imagery so the website feels premium and professional.
- [ ] As a host, I want to create destination pages so guests can explore locations and experiences.
- [ ] As a host, I want to create custom landing pages for marketing campaigns.
- [x] As a host, I want my website to automatically adapt to mobile devices.
- [x] As a host, I want to manage multiple websites for different brands or destinations.
- [ ] As a host, I want to add promotional banners and announcements to highlight offers.
- [ ] As a host, I want to create blog content to improve SEO and organic traffic.
- [x] As a host, I want to showcase luxury amenities visually to improve conversions.
- [ ] As a host, I want to display social proof and reviews on the homepage.

## Admin Stories

- [ ] As an admin, I want to manage template settings globally.
- [ ] As an admin, I want to enable or disable website modules.
- [ ] As an admin, I want to moderate website content before publishing.

---

# 🏡 2. Property Listing Management

## Host Stories

- [x] As a host, I want to add and edit property listings easily.
- [ ] As a host, I want to upload high-resolution galleries and videos.
- [ ] As a host, I want to categorize listings by property type.
- [x] As a host, I want to manage amenities and services per listing.
- [ ] As a host, I want to create multilingual property descriptions.
- [ ] As a host, I want to duplicate listings to speed up onboarding.
- [ ] As a host, I want to display nearby attractions and local guides.
[...]

PR #3172, opened 2026-09-04
#3038 Add quote panel to direct booking websites

karimh01:
## Close or Relate the Issue
- Closes #

- Related Issue: #3038

## Proposed Changes
Description:

Adds the panel where a guest on a direct booking website picks dates, sets the number of guests, and gets a real price back from the server.

This uses the endpoint from #3168, which is now live in API Gateway. Tested end to end on a published site.

What it does:
* The hero button now works. It scrolls down to the calendar instead of doing nothing
* The existing availability calendar becomes selectable: first click is check-in, second is check-out. Blocked days can end a stay but not start one, since checkout is exclusive
* A panel next to the calendar holds the guest counter and the button. The price comes back split out into nights, cleaning fee and total, with the time it stays valid
* Changing the dates or the guest count greys out the price with a note to check again, instead of clearing it
* Every error code from the endpoint gets its own message. Date problems show at the dates, guest problems at the counter, and anything else as a panel message with a retry button

Where it does and doesn't show:
The panel only renders on the published live site, never in the editor or the draft preview. On top of that a host has to switch it on with a new toggle, and it defaults to off, so nothing changes for existing sites until someone opts in.

Client-side date filtering is convenience only. The server decides what's available and what it costs.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Huge change (1000+ lines, mostly refactored/moved code. Clarify in [Refactoring](#Refactoring))
- [ ] Big change (+-max 1000)

## Refactoring

N/A. Nothing was moved or refactored, this is new code plus additions to existing files. The 65 deletions are lines that got replaced in place, mostly the visibility defaults map and the calendar call in the three templates.

## Evidence
- [x] Before/after images added when relevant

Quote panel on the live site with a real price from the endpoint:
image

Same panel on mobile:
image

Same panel on tablet:
image

## Responsiveness
- [x] UI checked on mobile
- [x] UI checked on tablet
- [x] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
  - [x} No global styling (see [#1691](https://github.com/domits1/Domits/issues/1691))
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

One thing worth flagging that came up while testing: you can publish a direct booking website for a listing that's still a draft. The quote then correctly returns a 422, but the guest just sees a generic "can't book right now". Might be worth an issue on the publish check.

## Keep or delete my branch
- [ ] Keep my branch

### DIFF
--- a/frontend/web/src/features/hostdashboard/website/WebsitePublicSitePage.jsx
+++ b/frontend/web/src/features/hostdashboard/website/WebsitePublicSitePage.jsx
@@ -362,6 +362,7 @@ function WebsitePublicSitePage() {
             showContactWidget={publicModel.visibility?.chatWidget ?? true}
             showBrowserChrome={false}
             enableScrollReveal={isPanoramaTemplate}
+            quoteContext={renderPayload?.site?.id ? { siteId: String(renderPayload.site.id) } : null}
           />
         </div>
       </main>
--- a/frontend/web/src/features/hostdashboard/website/rendering/AvailabilityCalendarPreview.jsx
+++ b/frontend/web/src/features/hostdashboard/website/rendering/AvailabilityCalendarPreview.jsx
@@ -89,9 +89,6 @@ const selectionPropType = PropTypes.shape({
   onSelectDate: PropTypes.func,
 });
 
-// Selection is opt-in: the calendar stays a display-only snapshot in the editor
-// and the draft preview, and becomes a date-range picker only when the booking
-// section on the live site hands it a selection.
 const resolveCellSelectionState = (cell, selection) => {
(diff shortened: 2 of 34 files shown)


BeMounir (approved):
everything looks good. i checked and also asked copilot and everything looks good. good test coverage aswell

elcoleta (approved):

PR #3175, opened 2026-09-04
#3038 Add direct booking website columns to the booking table

karimh01:
## Close or Relate the Issue
- Closes #

- Related Issue: #3038

## Proposed Changes
Description:

Adds five columns to the booking table so a booking from a direct booking website can be stored, and fixes the model and schema.psql where they didn't match the actual database anymore.

**What I did to the database (already applied and checked before opening this PR)**

First I read the live catalog instead of trusting the files, and found a few things:

- test.booking was missing four columns that main.booking has: refunded_amount, stripe_refund_id, refund_error and total_price. The refund migration from May says it did both schemas, but test never got them. So anything running with TEST=true against that schema has been failing for months
- total_price isn't in any migration file at all, for either schema
- that same migration says refunded_amount gets DEFAULT 0, but the live column has no default. So that never happened either

I fixed test first with a separate migration record, then added the five new columns to both schemas plus two unique indexes on public_booking_ref and idempotency_key. Both tables are now at 24 columns, all four indexes report indisvalid = t, and existing bookings still load fine in the app.

There's no migration runner in this repo, so I applied the SQL by hand through the DSQL runbook in the docs. Each migration has a .js record and a .sql file with the statements. The run order, the DSQL rules and what each verification query should return are in docs/internal/tools/dsql_booking_columns_runbook.md, since comments aren't allowed in code here.

**Code changes**

- Booking.js: the five columns, all nullable, no defaults. Merging this redeploys every Lambda, which is fine now because the columns already exist
- schema.psql: brought in line with the catalog for both schemas. Added the five new columns and the two indexes, plus total_price and bookingtype which were live but never written down. bookingtype is recorded as unbounded VARCHAR because that's what the catalog shows, and I removed DEFAULT 0 from refunded_amount since it isn't actually there
- bookingEntity.test.js: a test that fails if the entity's columns drift from the real table, or if schema.psql drifts from the entity. Figured that was worth adding given what I ran into

**About the columns.** NULL means it's a normal marketplace booking. booking_source will hold STANDALONE_SITE once the booking endpoint exists. I made everything VARCHAR(255) on purpose: DSQL has no ALTER COLUMN TYPE, so if a column turns out too narrow later you'd have to drop and recreate it on a production table.

**Something I found but didn't fix here.** Booking.js declares three defaults that the database doesn't have: total_price and refunded_amount both say 0, bookingtype says "direct". They don't break anything at runtime, but the model and the database disagree. I left it out so this PR stays about one thing, and put an it.todo in the new test as a reminder.

**Not in here:** making guestid nullable. DSQL can't add NOT NULL back once you drop it, so that's a decision I'd rather make together with whoever builds the endpoint that needs it.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [x] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
- N/A

## Evidence
No UI changes. The evidence is the verification queries in the .sql files: column counts before and after, a LEFT JOIN that proves both schemas match, pg_index.indisvalid on all four indexes, sys.jobs completed, and smoke inserts that rolled back cleanly. The doc lists what each one should return.

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch

### DIFF
--- a/backend/ORM/migrations/20260904_booking_direct_booking_website_columns.js
+++ b/backend/ORM/migrations/20260904_booking_direct_booking_website_columns.js
@@ -1,0 +1,141 @@
+export class BookingDirectBookingWebsiteColumns20260904 {
+  async up(queryRunner) {
+    await queryRunner.query(`
+      ALTER TABLE test.booking
+      ADD COLUMN IF NOT EXISTS booking_source VARCHAR(255);
+    `);
+
+    await queryRunner.query(`
+      ALTER TABLE test.booking
+      ADD COLUMN IF NOT EXISTS site_id VARCHAR(255);
+    `);
+
+    await queryRunner.query(`
+      ALTER TABLE test.booking
+      ADD COLUMN IF NOT EXISTS guest_email VARCHAR(255);
+    `);
+
+    await queryRunner.query(`
+      ALTER TABLE test.booking
+      ADD COLUMN IF NOT EXISTS public_booking_ref VARCHAR(255);
+    `);
+
+    await queryRunner.query(`
+      ALTER TABLE test.booking
+      ADD COLUMN IF NOT EXISTS idempotency_key VARCHAR(255);
+    `);
+
+    await queryRunner.query(`
+      CREATE UNIQUE INDEX ASYNC booking_public_booking_ref_unique_test
+      ON test.booking (public_booking_ref);
+    `);
+
+    await queryRunner.query(`
+      CREATE UNIQUE INDEX ASYNC booking_idempotency_key_unique_test
+      ON test.booking (idempotency_key);
+    `);
+
+    await queryRunner.query(`
+      ALTER TABLE main.booking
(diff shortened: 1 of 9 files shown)


BeMounir (commented):
sonar is reading alot of dublicate code and is getting really angry because of it

BeMounir (approved):
while sonar did pass it. it is still running into some errors. mostly just duplicates. other than that it does look good

EnesVeli (approved):
Nice work!

Issue #2266, opened 2025-11-13
Use Case / Scenario Testing CM (Channel Manager)


stefanhopman1:
Channel Management System for short-term rentals is a system that synchronizes listings, rates, and availability across multiple online travel agencies (OTAs) such as Airbnb, Booking.com, VRBO, Expedia, etc.

----

Use Case / Scenario Testing: [PMS](https://github.com/domits1/Domits/issues/2262), [CRS](https://github.com/domits1/Domits/issues/2265), [CM](https://github.com/domits1/Domits/issues/2266), [RMS](https://github.com/domits1/Domits/issues/2267), [Calendar & Prices](https://github.com/domits1/Domits/issues/2299), [Inbox](https://github.com/domits1/Domits/issues/2291), [Finance](https://github.com/domits1/Domits/issues/2287)
Main issues: [PMS](https://github.com/domits1/Domits/issues/149), [CRS](https://github.com/domits1/Domits/issues/501), [CM](https://github.com/domits1/Domits/issues/440), [RMS](https://github.com/domits1/Domits/issues/436), [Calendar & Prices](https://github.com/domits1/Domits/issues/36), [Inbox](https://github.com/domits1/Domits/issues/767), [Finance](https://github.com/domits1/Domits/issues/163)

----

👥 User Roles

- Host / Property Manager
- Channel Manager / Revenue Manager
- Admin / System Operator
- Cleaner / Maintenance Staff
- Accountant / Finance Manager
- OTA System (external API partner)

🏘️ 1. Property Listing Management

- [ ] As a host, I want to connect my properties to multiple OTAs (Airbnb, Booking.com, VRBO) so I can manage them centrally.
- [ ] As a host, I want to import existing listings from OTAs into the system to avoid duplicate data entry.
- [ ] As a host, I want to push property details (photos, descriptions, amenities) to all connected channels.
- [ ] As a host, I want to update listing information (description, amenities, photos) once and sync it everywhere.
- [ ] As a host, I want to unpublish a property from specific channels temporarily.
- [ ] As a host, I want to group properties by portfolio, city, or client for easier management.
- [ ] As a host, I want to view which channels each property is connected to.
- [ ] As an admin, I want to validate property data before pushing it to external APIs.

📅 2. Availability & Calendar Synchronization

- [ ] As a host, I want my property calendar to sync across all OTAs automatically.
- [ ] As a host, I want bookings made on any channel to instantly block dates on all others to prevent double bookings.
[...]

PR #3180, opened 2026-09-07
test(cm): add conflict detection and ical sync scenario tests #2266

EnesVeli:
## Close or Relate the Issue
- Closes #

- Related Issue: #2266 

## Proposed Changes
Description:
Adds scenario test coverage for double-booking and calendar-conflict prevention, anchored to Issue 
[#2266](https://github.com/domits1/Domits/issues/2266)
 acceptance criteria section 2 ("Bookings on any channel instantly block dates on all others"; "Detect and resolve calendar conflicts automatically") and section 4 ("Differentiate between confirmed, pending, and canceled bookings"; "Modify or cancel bookings without self-conflict"; "Detect duplicate reservations and flag them"). No production code changed — tests only, covering three previously-untested/under-tested paths:

reservationRepository.conflict.test.js — same-channel double-booking guard (ReservationRepository.assertNoBookingConflict). Covers blocking vs. non-blocking booking statuses, duplicate-reservation detection, excludeBookingId self-conflict exclusion, property-scoping of the conflict query, and the check-in/out buffer gap.
externalCalendarService.conflict.test.js — cross-channel (iCal) double-booking guard (ExternalCalendarService.ensureNoExternalConflict). Covers no-sources pass-through, non-overlapping vs. overlapping blocked dates, aggregation across multiple external sources, malformed-source skipping, and ServiceUnavailableException resilience when a calendar fetch fails.
service.conflict.test.js — the Ical-sync-scheduler background job that keeps the iCal blocked-dates cache fresh for (2) above. First-ever test coverage for this function (no test/Ical-sync-scheduler directory existed before). Covers due/not-due sync scheduling, the force override, partial-failure resilience (one broken source doesn't block the batch), and the same malformed-source guard.

All three files went through the eng-review skill gate before commit; each found (and had fixed) a coverage gap around asserting the underlying repository write actually happens on the success path, not just the reject/resolve outcome.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size

## Refactoring
- src/features/example.xx

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [ x] PR title is descriptive and includes issue number
- [ x] Conventions
- [x ] Testing
  - [x ] Code tested locally
  - [ x] Jest tests are included
  - [ x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch

### DIFF
--- a/backend/test/General-Bookings-CRUD-Bookings-develop/externalCalendarService.conflict.test.js
+++ b/backend/test/General-Bookings-CRUD-Bookings-develop/externalCalendarService.conflict.test.js
@@ -1,0 +1,125 @@
+const mockListSources = jest.fn();
+const mockUpsertSource = jest.fn();
+jest.mock("../../functions/General-Bookings-CRUD-Bookings-develop/data/externalCalendarRepository.js", () => ({
+  __esModule: true,
+  default: jest.fn().mockImplementation(() => ({
+    listSources: mockListSources,
+    upsertSource: mockUpsertSource,
+  })),
+}));
+
+const mockFetchExternalCalendar = jest.fn();
+jest.mock("../../functions/.shared/icalTransport.js", () => ({
+  __esModule: true,
+  buildSourceUpsertPayload: jest.fn((args) => args),
+  fetchExternalCalendar: (...args) => mockFetchExternalCalendar(...args),
+}));
+
+const ExternalCalendarService =
+  require("../../functions/General-Bookings-CRUD-Bookings-develop/business/externalCalendarService.js").default;
+const ConflictException =
+  require("../../functions/General-Bookings-CRUD-Bookings-develop/util/exception/ConflictException.js").default;
+const ServiceUnavailableException =
+  require("../../functions/General-Bookings-CRUD-Bookings-develop/util/exception/ServiceUnavailableException.js").default;
+
+const BASE_RANGE = {
+  arrivalMs: Date.parse("2026-06-15T00:00:00.000Z"),
+  departureMs: Date.parse("2026-06-17T00:00:00.000Z"),
+};
+
+const buildEvent = (dtstart, dtend) => ({ Dtstart: dtstart, Dtend: dtend });
+
+describe("ExternalCalendarService cross-channel double-booking guard", () => {
+  let consoleErrorSpy;
+
+  beforeEach(() => {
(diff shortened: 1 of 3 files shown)


karimh01 (approved):
Looks good, fine by me.

Good that you check the query itself instead of just the result, especially in assertNoBookingConflict. And nice that the excludedStatuses are pinned down now, including both spellings of Cancelled.

Two things:

In reservationRepository.conflict.test.js, the case "an overlapping booking in a non-blocking status" mocks getCount to 0 and then checks nothing throws. That only proves we don't throw on zero, not that cancelled bookings get filtered out. You do cover that in the separate test below, so it's fine, the name just says more than the test does.

And in the scheduler, a source without a calendarUrl gets skipped but still counts as succeeded. Is that on purpose? Seems like you'd want to see somewhere that a source is misconfigured.

EnesVeli:
> Looks good, fine by me.
> 
> Good that you check the query itself instead of just the result, especially in assertNoBookingConflict. And nice that the excludedStatuses are pinned down now, including both spellings of Cancelled.
> 
> Two things:
> 
> In reservationRepository.conflict.test.js, the case "an overlapping booking in a non-blocking status" mocks getCount to 0 and then checks nothing throws. That only proves we don't throw on zero, not that cancelled bookings get filtered out. You do cover that in the separate test below, so it's fine, the name just says more than the test does.
> 
> And in the scheduler, a source without a calendarUrl gets skipped but still counts as succeeded. Is that on purpose? Seems like you'd want to see somewhere that a source is misconfigured.


Thanks for both fixed the test name renamed to resolves when the query
reports no conflicting bookings, since that's actually what that row checks;
the real proof that non-blocking statuses get filtered is the dedicated
query-clause test below it.

On the scheduler: good catch, and no, it's not intentional. A misconfigured
source just returns quietly instead of throwing, so it ends up counted the
same as a real success. Filed it separately #3199 since fixing it
properly means changing what the sync result actually reports, which felt
like the wrong thing to sneak into a tests-only PR.

BeMounir (approved):

PR #3176, opened 2026-09-04
test(cm): add ARI scenario and resilience tests in PropertyHandler #2266

EnesVeli:
## Close or Relate the Issue
[//]: # (Only if this should close the issue)
- Closes #

[//]: # (in the issue a reference will be added to this pr)
- Related Issue: #2266 

## Proposed Changes
[//]: # (Add a detailed description of the changes here)
Description:
Adds scenario tests for `ChannexCalendarChangeSyncClient` covering rate changes, availability blocks, stay restrictions, and combined multi-type changes. Also adds resilience coverage for failure paths that were previously untested: Lambda invocation throwing (timeout/network error), Lambda `FunctionError`, and a missing/empty response body on a 2xx status.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [x] Optimization
- [ ] Documentation update

## Change Size
[//]: # (Indicate the size of this change.)
[//]: # (Clarify under Refactoring which parts are moved/unchanged code, so the reviewer doesn’t review old logic.)
- [ ] Huge change (1000+ lines, mostly refactored/moved code. Clarify in [Refactoring](#Refactoring))
- [ ] Big change (+-max 1000)
- [x] Small change (less than 300)

## Refactoring
Refactors following files, but didn't change the code. This is to avoid reviewing old logic.
N/A — no existing code was moved or refactored, only new test cases added.

## Evidence
- [ ] Screenshots or screen recording added for UI changes
- [ ] Before/after images added when relevant
- [ ] Images clearly show the implemented change
- [ ] Image quality is readable (no cropped or blurry evidence)

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop
- [ ] No broken layout, overflow, or hidden content on tested screen sizes

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated
- [ ] Checked for vulnerabilities using "npm audit"

## Checklist
[//]: # (All boxes must be checked before merging.)
- [ x] Pull request has at least 2 reviewers assigned
- [x ] PR title is descriptive and includes issue number
- [x ] Conventions
  - [x ] No config files have been added (Amplify config, cache files)
  - [x ] No hardcoded sensitive data (e.g., API-keys/passwords)
  - [x ] No global styling (see [#1691](https://github.com/domits1/Domits/issues/1691))
  - [ x] No `console.log` left in code
  - [x ] No commented-out code remains
- [ x] Testing
[...]

PR #3181, opened 2026-09-07
test(cm): add inquiry conflict auto-resolution scenario tests #2266

EnesVeli:
## Close or Relate the Issue
- Closes #

- Related Issue: 

## Proposed Changes
Description:
Adds scenario test coverage for inquiry conflict auto-resolution — Group A of the #2266 test-coverage effort (see the prior PR covering same-channel/cross-channel double-booking guards + the iCal sync scheduler). This is the last previously-untested core conflict primitive: accepting one inquiry auto-declines every competing inquiry on the same property/dates, which is the direct implementation of AC section 2's "detect and resolve calendar conflicts automatically." No production code changed — tests only.

bookingService.inquiry.test.js — BookingService.acceptInquiry/declineInquiry. Covers the shared guard clauses (not-found, host-only Forbidden, wrong-status rejection) via a parametrized table across both methods, the auto-decline fan-out with declinedCount reporting, self-exclusion from the overlap search, the zero-overlap case, and declineInquiry's narrower blast radius (only touches the declined booking, never searches for overlaps). Also includes a test pinning acceptInquiry's existing partial-failure window: the accept commits to Awaiting Payment before the overlap search runs, so if that search fails, the accepted booking stays live while nothing else changes — documented as current behavior, not fixed here (that would be a production change, out of scope for a tests-only PR).
reservationRepository.overlappingInquiries.test.js — ReservationRepository.getOverlappingInquiries query-builder logic. Covers each query clause (property scoping, Inquiry-only status filter, self-exclusion, arrival/departure overlap bounds) and confirms the returned rows pass through unmodified. Notably, unlike assertNoBookingConflict from the prior PR, this query has no check-in/out buffer gap — documented explicitly in a comment so a future reader doesn't assume the asymmetry is a bug.

Verified with a live mutation test (temporarily disabling the auto-decline fan-out in production code, confirming the corresponding test fails, then restoring) that the core assertion isn't just green-by-accident. Went through the eng-review skill gate before commit; findings applied — the not-found mock now matches production's real { message, statusCode: 204 } return shape, and the partial-failure window is now pinned by a test.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
- src/features/example.xx

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x]Pull request has at least 2 reviewers assigned
- [x ] Conventions
- [x ] Testing
  - [x ] Code tested locally
  - [ x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch

### DIFF
--- a/backend/test/General-Bookings-CRUD-Bookings-develop/bookingService.inquiry.test.js
+++ b/backend/test/General-Bookings-CRUD-Bookings-develop/bookingService.inquiry.test.js
@@ -1,0 +1,161 @@
+jest.mock("@aws-sdk/client-lambda", () => ({
+  LambdaClient: jest.fn().mockImplementation(() => ({ send: jest.fn() })),
+  InvokeCommand: jest.fn().mockImplementation((input) => input),
+}));
+
+const BookingService =
+  require("../../functions/General-Bookings-CRUD-Bookings-develop/business/bookingService.js").default;
+const Forbidden = require("../../functions/General-Bookings-CRUD-Bookings-develop/util/exception/Forbidden.js").default;
+const NotFoundException =
+  require("../../functions/General-Bookings-CRUD-Bookings-develop/util/exception/NotFoundException.js").default;
+const {
+  BadRequestException,
+} = require("../../functions/General-Bookings-CRUD-Bookings-develop/util/exception/badRequestException.js");
+
+const HOST_ID = "host-1";
+const INQUIRY_ID = "inquiry-1";
+
+const buildBooking = (overrides = {}) => ({
+  id: INQUIRY_ID,
+  hostid: HOST_ID,
+  property_id: "property-1",
+  status: "Inquiry",
+  arrivaldate: Date.parse("2026-06-15T00:00:00.000Z"),
+  departuredate: Date.parse("2026-06-17T00:00:00.000Z"),
+  ...overrides,
+});
+
+const buildService = ({ booking = buildBooking(), overlapping = [], userSub = HOST_ID } = {}) => {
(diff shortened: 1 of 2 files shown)


karimh01 (approved):
Looks good, fine by me.

Good improvement over the last one: the it.each now checks each clause with its exact parameters instead of mocking a count and asserting on that.

And the test about the overlap search failing after the accept is committed is a useful one to have. It doesn't fix the gap but at least it's written down now.

Same thing as #3176 though: there are comments in the test files. Robert asked me on Friday to strip all comments from code, test files included. So you'll probably get the same question.

EnesVeli:
> Looks good, fine by me.
> 
> Good improvement over the last one: the it.each now checks each clause with its exact parameters instead of mocking a count and asserting on that.
> 
> And the test about the overlap search failing after the accept is committed is a useful one to have. It doesn't fix the gap but at least it's written down now.
> 
> Same thing as #3176 though: there are comments in the test files. Robert asked me on Friday to strip all comments from code, test files included. So you'll probably get the same question.

Thanks. stripped both comments. Folded the "no check-in/out buffer" detail into
the clause descriptions themselves instead of a comment, so it's still visible
in the test output, not just gone.

elcoleta (approved):

PR #3170, opened 2026-09-03
#3038 Add public quote endpoint for direct booking websites

karimh01:
## Close or Relate the Issue
[//]: # (Only if this should close the issue)
- Closes 

[//]: # (in the issue a reference will be added to this pr)
- Related Issue: #3038

## Proposed Changes
[//]: # (Add a detailed description of the changes here)
Description:

Adds an endpoint so a guest on a direct booking website can ask what a stay costs: POST /property/website/public/quote.

This builds on #3168, which added the code that does the actual work (checking the dates, checking if it's still free, calculating the price). This PR just makes that reachable from the website. That's why it targets #3168 instead of acceptance; once that one merges, the base moves to acceptance by itself.

Why everything happens on the server: guests on these sites don't have an account, so there's no login to check. If we calculated the price in the browser, anyone could open devtools and change it. So the server decides everything and the browser is trusted with nothing.

What it does:
* Takes only the siteId from the request and looks up the property in the database. If the browser sends a propertyId, it gets ignored, so you can't ask for a price on someone else's property
* Handles three cases for the site: 404 if it doesn't exist, 410 if it's suspended, 409 if it isn't published yet. In all three, no price is returned and nothing is logged
* Turns the error codes from the quote engine into HTTP status codes
* Writes three events for the KPI dashboard. The conflict event only fires when the dates are actually taken, not when someone types a wrong date, so input mistakes don't count as conflicts
* Returns a generic 500 on unexpected errors, so nothing internal ends up in the response

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
[//]: # (Indicate the size of this change.)
[//]: # (Clarify under Refactoring which parts are moved/unchanged code, so the reviewer doesn’t review old logic.)
- [ ] Huge change (1000+ lines, mostly refactored/moved code. Clarify in [Refactoring](#Refactoring))
- [x] Big change (+-max 1000)
- [ ] Small change (less than 300)

## Refactoring
Refactors following files, but didn't change the code. This is to avoid reviewing old logic.
- N/A

## Evidence
- [ ] Screenshots or screen recording added for UI changes
- [ ] Before/after images added when relevant
- [ ] Images clearly show the implemented change
- [ ] Image quality is readable (no cropped or blurry evidence)

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop
[...]

PR #3184, opened 2026-09-07
#3038 Add booking request endpoint for direct booking websites

karimh01:
Reopened because #3178 closed automatically when I deleted the base branch. Same code, now targeting acceptance directly.

# Close or Relate the Issue
- Closes #

- Related Issue: #3038

## Proposed Changes
Description:

Adds the endpoint where a guest on a direct booking website actually sends in a booking request: `POST /public/sites/{siteId}/bookings`. Builds on #3175, so it points at that branch until it merges.

The guest has no account. They send the token they got when they asked for a price (#3170), plus their name and email.

**What happens when someone sends a request**

First it checks the Idempotency-Key header and whether the name and email look valid. Nothing else runs until that's fine.

Then it checks the token. The dates, guest count and price all come from the token, not from what the browser sends. Otherwise you could just make up your own price.

Then it looks up the idempotency key. If someone double clicks, the same key comes in twice and they get the same answer back instead of two bookings. Same key with different dates gets rejected.

Then it loads the site, checks it's published, and checks the site actually owns the property the token was made for.

Then it asks the quote endpoint for a fresh price. Prices can change between asking and booking, so the token only proves what was quoted, not what it costs now. If the total is different the guest has to check again.

Then it runs the same availability checks the normal booking path runs, saves the booking as an Inquiry, records an event and sends two emails. Those last two are best effort — the booking is already saved, so a failed email shouldn't turn a 201 into an error.

**Stuff worth knowing**

`guestid` can't be empty and there's no account, so the row gets a generated `dbw-<uuid>`. Not one fixed value for everyone, because then every anonymous guest would share the same identity. Anything that reads `guestid` as an account id needs to check `booking_source` first.

`hostname` now gets the host's real name, and falls back to the old `"WIP-Host"` value if the lookup fails. That lookup already happened for the host email, so it only meant moving one call.

`total_price` is stored in euros, same as the Stripe path does. The API gives back cents.

`HostReservationDetails` now reads `guest_email` from the booking, so a host can actually reply. Without it they'd get a request they can't answer.

**What doesn't work yet, so don't merge this thinking it's done**

The guest never hears the outcome. Accept and decline don't notify anyone on the guest side, and these guests have no dashboard. That's for the payment story.

The guest can't pay. After the host accepts it sits in Awaiting Payment, but the way to Stripe needs an account the guest doesn't have. Also the payment story.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Huge change ((1000+ lines, mostly refactored/moved code. Clarify in [Refactoring](#Refactoring))
- [ ] Big change (+-max 1000)

## Refactoring
Nothing refactored or moved. The 1959 additions are all new: roughly half is tests (nine suites, 84 tests), the rest is the new endpoint split over util, data, business and controller. Only 4 deletions, and one function in `propertyRepository.js` went from private to exported so the new repositories can reuse it instead of copying the same logic.

## Evidence
No UI changes apart from one line in HostReservationDetails. The endpoint has no route in API Gateway yet so I can't try it end to end — the tests are the evidence.

## Responsiveness
Backend, plus one line of frontend that doesn't change any layout.

- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


BeMounir (approved):
PR was already approved but got deleted. Approving it again.

elcoleta (approved):

PR #3187, opened 2026-09-08
#3038 Add booking request form to direct booking websites

karimh01:
## Close or Relate the Issue
- Closes #

- Related Issue: #3038

## Proposed Changes
Description:

Adds the form where a guest actually sends in a booking request. After they get a live price, the panel asks for their name and email and posts to POST /public/sites/{siteId}/bookings with the quote token they already have.

**How it works**

The guest picks dates, hits check availability, gets a price. Below that price a short form appears with two fields and a submit button. The form only shows when the quote is fresh, so if they change the dates or the 30 minutes run out, it disappears and they have to check again first. That way the token being sent is always the one whose price they're looking at.

Every submit carries an Idempotency-Key header. That key is stored per site and per quote, so if someone clicks twice or the connection drops and they retry, the server sends back the original booking instead of making a second one. On success the key gets cleared.

The success screen shows the booking reference and says the host still has to confirm. Nothing is booked yet at that point, and the wording is explicit about that so nobody books a flight on it.

**About the errors**

This PR keeps error handling simple on purpose. A rejected name or email shows on the field, a site that can't take bookings hides the whole thing, and everything else gets one message with a retry: "We couldn't send your request. Please try again." The retry is safe because it reuses the same key.

The per-code handling (price changed, dates taken, and so on) is the next PR, together with turning the panel on by default. That way this one reviews the flow and that one reviews what each failure looks like.

**Visibility stays off**

The booking panel default is still false, so nothing changes for existing sites. Only hosts who already turned the panel on themselves get the form.

**Dependencies**

The endpoint from #3184 is live and reachable, with Idempotency-Key in the CORS allowed headers and throttling in place. Verified with curl from outside before opening this.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Huge change (1000+ lines, mostly refactored/moved code. Clarify in [Refactoring](#Refactoring))
- [ ] Big change (+-max 1000)

## Refactoring
Nothing refactored or moved. It's 1546 lines of new code, roughly half of it tests (7 of the 17 files). The service, the hook and the error copy follow the same shape as the existing quote versions, but they're written fresh, not copied.

## Evidence
The form and success view are new UI in the booking panel. I can add screenshots if useful, but you need a published site with the panel turned on to see the flow.

## Responsiveness
The form sits in the existing panel and uses its layout, so it follows whatever the panel already does.

- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [x] Keep my branch

### DIFF
--- a/frontend/web/src/features/hostdashboard/website/rendering/booking/BookingRequestForm.jsx
+++ b/frontend/web/src/features/hostdashboard/website/rendering/booking/BookingRequestForm.jsx
@@ -1,0 +1,108 @@
+import React, { useId } from "react";
+import PropTypes from "prop-types";
+import styles from "./QuoteAvailabilitySection.module.scss";
+
+function ContactField({ id, label, type, autoComplete, value, error, disabled, onChange }) {
+  const errorId = `${id}-error`;
+
+  return (
+    <div className={styles.formField}>
+      <label htmlFor={id} className={styles.fieldLabel}>
+        {label}
+      </label>
+      <input
+        id={id}
+        className={`${styles.input} ${error ? styles.inputInvalid : ""}`.trim()}
+        type={type}
+        autoComplete={autoComplete}
+        value={value}
+        disabled={disabled}
+        aria-invalid={error ? "true" : undefined}
+        aria-describedby={error ? errorId : undefined}
+        onChange={(event) => onChange(event.target.value)}
+      />
+      {error ? (
+        <p id={errorId} className={styles.fieldError}>
+          {error}
+        </p>
+      ) : null}
+    </div>
+  );
+}
+
+ContactField.propTypes = {
+  id: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
+  type: PropTypes.string.isRequired,
+  autoComplete: PropTypes.string.isRequired,
+  value: PropTypes.string.isRequired,
+  error: PropTypes.string,
+  disabled: PropTypes.bool.isRequired,
+  onChange: PropTypes.func.isRequired,
+};
+
+export default function BookingRequestForm({
+  guest,
+  onGuestChange,
+  onSubmit,
+  isSubmitting = false,
+  fieldErrors = {},
+  formError = "",
+}) {
+  const fieldIdPrefix = useId();
+
+  return (
+    <form
+      className={styles.form}
+      noValidate
+      onSubmit={(event) => {
+        event.preventDefault();
+        onSubmit();
+      }}>
+      <ContactField
+        id={`${fieldIdPrefix}-name`}
+        label="Your name"
+        type="text"
+        autoComplete="name"
+        value={guest?.name || ""}
+        error={fieldErrors?.name || ""}
+        disabled={isSubmitting}
+        onChange={(value) => onGuestChange("name", value)}
+      />
+      <ContactField
+        id={`${fieldIdPrefix}-email`}
+        label="Email address"
+        type="email"
+        autoComplete="email"
+        value={guest?.email || ""}
+        error={fieldErrors?.email || ""}
+        disabled={isSubmitting}
+        onChange={(value) => onGuestChange("email", value)}
+      />
+      {formError ? (
+        <p className={styles.fieldError} role="alert">
+          {formError}
+        </p>
+      ) : null}
+      <button type="submit" className={styles.action} disabled={isSubmitting}>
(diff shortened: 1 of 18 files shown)


elcoleta (approved):
Looks good overall. The flow is well scoped, the quote freshness/idempotency handling is thoughtful, and the tests cover the main request path and failure states clearly.

Mehedihasanb (changes requested):
The booking-request flow looks well thought out overall. The fresh-quote gating, guest validation, success state, request payload, and idempotency handling all have good test coverage.

I do see one blocker before approval: SonarCloud currently reports Cognitive Complexity 30 for QuotePanel, above the allowed 15. Since this complexity was introduced by this PR, I think QuotePanel should be split/refactored so the quality gate passes before merge.

One additional question around the idempotency handling: when the server returns `idempotency_key_reused`, the hook clears the existing key and automatically retries with a new one. Can we confirm that this response guarantees no booking was created for the previous key? Since this is a booking-creation endpoint, I'd want to make sure rotating the key cannot ever turn an ambiguous retry into a duplicate booking.

Otherwise the flow looks good to me.

karimh01:
> The booking-request flow looks well thought out overall. The fresh-quote gating, guest validation, success state, request payload, and idempotency handling all have good test coverage.
> 
> I do see one blocker before approval: SonarCloud currently reports Cognitive Complexity 30 for QuotePanel, above the allowed 15. Since this complexity was introduced by this PR, I think QuotePanel should be split/refactored so the quality gate passes before merge.
> 
> One additional question around the idempotency handling: when the server returns `idempotency_key_reused`, the hook clears the existing key and automatically retries with a new one. Can we confirm that this response guarantees no booking was created for the previous key? Since this is a booking-creation endpoint, I'd want to make sure rotating the key cannot ever turn an ambiguous retry into a duplicate booking.
> 
> Otherwise the flow looks good to me.

Split QuotePanel, complexity is down from 32 to 4 and nothing in the file is above 12 now. Pulled the two view derivations out into helpers and moved the stay summary and the panel body into their own components.

Render output is identical, all 32 existing tests pass without any changes, which is how I checked I didn't break anything.

EnesVeli (approved):
Nice work on the split — 32 down to 4, and identical render output verified by the existing tests is exactly the right way to check a refactor like that.

The other half of Mehedihasan's review is still open though, Their instinct was right, and the answer is the opposite of what the retry assumes: idempotency_key_reused guarantees a booking was created under the previous key.

In publicSiteBookingRequestService.js, that error is thrown at :373, and the only way to reach it is from :348 — the unique-violation branch on the idempotency key constraint. Reaching that branch means a row already exists for that key. If the stay matches, the server replays the original booking. The error is thrown only when a booking exists and it's for a different stay.

So in useWebsiteBookingRequest.js:53-59 (unchanged by the refactor):

if (error?.code !== IDEMPOTENCY_KEY_REUSED_CODE) throw error;
clearBookingIdempotencyKey(siteId);
result = await send(getOrCreateBookingIdempotencyKey({ siteId, quoteId: quote.quoteId }));

we drop the key and resend with a fresh one while the first booking still exists — creating a second. The one error meaning "you already booked" is the one we answer by booking again.

On how reachable it is, to be fair about severity: reuse is gated on record.quoteId === quoteId, so normally a date change produces a new quote, new quoteId, new key, and this never fires. The gap is cleanText(undefined) returning "", and "" === "" — if a quote ever comes back without a quoteId, every stay on that site shares one key. Then: book dates X → switch to dates Y → server returns idempotency_key_reused → we rotate → two bookings, dates X still held. Every test passes a real quoteId, so this is latent rather than routine, but the mechanism is live.

Suggested fix, both small:

Treat idempotency_key_reused as terminal rather than retryable — surface it and stop, since a booking provably exists. If you want recovery, fetch the existing booking instead of minting a new key.
Have isUsableRecord reject an empty quoteId, so a missing id can't alias two different stays onto one key.

Rest of the PR looks good to me — the quote-freshness gating is the right call, abort handling is correct, and the coverage is solid.

Mehedihasanb (approved):
Approved!!

PR #3188, opened 2026-09-08
#3038 Handle booking request failures and enable the booking panel

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3038

## Proposed Changes
Description:

Builds on the previous PR, which added the form with one generic error message. This one makes each failure say what actually went wrong, and turns the booking panel on by default.

**What each error does now**

If the price changed while the guest was filling in the form, we re-quote automatically and keep their details. They see the new price and can send again.

If the dates got taken in the meantime, the selection clears and those nights get blocked on the calendar, so they can't pick them again in that session.

If the dates don't meet the property's stay rules, the selection clears with the message from the server.

If the token is rejected, the quote is marked stale and they have to check availability again.

If the property can't be booked online, we show the host's WhatsApp link instead.

Everything else still gets a retry with the same key, so it's safe to press again.

**The calendar freezes while sending**

Found this while testing: the calendar stayed clickable during the second the request is in flight. Click another date in that window and the summary shows dates that weren't booked. Now the calendar is frozen while sending, and the confirmation reads its dates from the server response instead of from what's selected in the browser.

**Turning the panel on**

The visibility default goes from false to true. Visibility flags only get saved when they differ from the default, so every host who never touched the toggle picks this up on their next render, without republishing.

That's why it's in this PR and not the previous one: from the moment the panel is on for everyone, guests hit these failure paths without a host having opted in. So the error handling needs to be there first.

One edge worth knowing: a host who turned the panel on and then off again since #3172 also has nothing saved, since off was the default. They'll come up on. The toggle has only existed for a few days, so it's unlikely, but worth mentioning.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
Nothing refactored or moved.

## Evidence
Behaviour changes rather than layout. The failure paths need a published site with the panel on, plus dates that get taken between quoting and submitting, so they're covered by tests rather than screenshots.

## Responsiveness
No layout changes. The messages render in slots the panel already has.

- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages

- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included — one per recovery path, plus a test that the calendar is frozen while sending and the confirmation shows the server's dates
  - [x] Jest tests are passing — 14 suites, 138 tests in the website directory

## Keep or delete my branch
- [ ] Keep my branch


karimh01:
Depends on #3187, that one needs to go in first. This PR targets that branch, so the diff here only shows the error handling and the visibility flag.

EnesVeli (approved):
Nice work verified every error code this PR maps against PublicBookingRequestError.js on the backend, every one is real, so the mapping actually fires in production rather than silently falling through. The calendar-freeze fix is correctly targeted too: isSelectionFrozen blocks selection during the in-flight window, and the confirmation now reads dates from the server result instead of live range, which closes exactly the race described.

One blocker: QuotePanel.jsx:170-171 references styles.notice and styles.noticeMessage, but neither class exists in QuoteAvailabilitySection.module.scss  this PR doesn't touch that file. The new re-quote notice (quote_expired path) will render unstyled, next to bookingPanelError's properly-styled .alert/.alertMessage banner right below it. Since this PR also flips quotePanel to visible-by-default, this ships to real guests immediately on merge. Easy fix: either reuse .alert/.alertMessage, or add .notice/.noticeMessage rules to the scss file.

One non-blocking gap: idempotency_key_reused (bookingRequestErrorCopy.js:39-95) has no explicit branch, so it falls to the generic catch-all "Something went wrong... Please try again." with recovery: RETRY. That code only fires when a booking already exists under that key for a different stay, so "try again" is actively wrong there, and the Retry button would just repeat the same rotate-and-resend from #3187's hook, risking another duplicate booking. Worth its own branch — PANEL scope, recovery: NONE, something like "a request for these dates may already be on its way — check your email before trying again." Since this PR is specifically about precise per-code messaging, it's a natural fit, but it's a copy gap tied to #3187's rotation behavior rather than a new defect here, so not blocking. Happy to file it as a quick follow-up issue if you'd rather keep it out of this PR.

Approving believing that you will check on my comment, and fix up the small issue if needed,

karimh01:
> Nice work verified every error code this PR maps against PublicBookingRequestError.js on the backend, every one is real, so the mapping actually fires in production rather than silently falling through. The calendar-freeze fix is correctly targeted too: isSelectionFrozen blocks selection during the in-flight window, and the confirmation now reads dates from the server result instead of live range, which closes exactly the race described.
> 
> One blocker: QuotePanel.jsx:170-171 references styles.notice and styles.noticeMessage, but neither class exists in QuoteAvailabilitySection.module.scss this PR doesn't touch that file. The new re-quote notice (quote_expired path) will render unstyled, next to bookingPanelError's properly-styled .alert/.alertMessage banner right below it. Since this PR also flips quotePanel to visible-by-default, this ships to real guests immediately on merge. Easy fix: either reuse .alert/.alertMessage, or add .notice/.noticeMessage rules to the scss file.
> 
> One non-blocking gap: idempotency_key_reused (bookingRequestErrorCopy.js:39-95) has no explicit branch, so it falls to the generic catch-all "Something went wrong... Please try again." with recovery: RETRY. That code only fires when a booking already exists under that key for a different stay, so "try again" is actively wrong there, and the Retry button would just repeat the same rotate-and-resend from #3187's hook, risking another duplicate booking. Worth its own branch — PANEL scope, recovery: NONE, something like "a request for these dates may already be on its way — check your email before trying again." Since this PR is specifically about precise per-code messaging, it's a natural fit, but it's a copy gap tied to #3187's rotation behavior rather than a new defect here, so not blocking. Happy to file it as a quick follow-up issue if you'd rather keep it out of this PR.
> 
> Approving believing that you will check on my comment, and fix up the small issue if needed,

Thanks for checking the error codes against the backend.

On the styling: .notice and .noticeMessage do exist, they're at line 247 and 255 in QuoteAvailabilitySection.module.scss. They came in with #3187, which is already merged, so this PR doesn't need to touch the file. Probably looks missing because you were reading the diff rather than the file on the branch.

The idempotency_key_reused copy is a fair point though. It falls to the generic retry message and "try again" is wrong for a code that means something already exists. I'd rather keep it out of this PR since it's tied to #3187's rotation behaviour, so yes please file it as a follow-up and I'll pick it up.

elcoleta (changes requested):
There’s still a similar race while the quote itself is being fetched. The guest count is disabled during QUOTE_STATUS.LOADING, but the calendar remains selectable because isSelectionFrozen only covers booking submission/success. If the guest changes the dates before the quote request finishes, notifySelectionChanged() marks the quote stale, but the in-flight request can still resolve afterwards and set it back to SUCCESS for the previous dates. That can leave the current calendar selection paired with a quote token for a different stay.
Since this PR enables the booking panel by default, I think we should also freeze calendar selection while the quote request is loading, or abort the in-flight quote when the selection changes.

karimh01:
> There’s still a similar race while the quote itself is being fetched. The guest count is disabled during QUOTE_STATUS.LOADING, but the calendar remains selectable because isSelectionFrozen only covers booking submission/success. If the guest changes the dates before the quote request finishes, notifySelectionChanged() marks the quote stale, but the in-flight request can still resolve afterwards and set it back to SUCCESS for the previous dates. That can leave the current calendar selection paired with a quote token for a different stay. Since this PR enables the booking panel by default, I think we should also freeze calendar selection while the quote request is loading, or abort the in-flight quote when the selection changes.

Fixed both ways. notifySelectionChanged now aborts the in-flight request, and the section checks the quote's dates and guest count against the current selection before treating it as usable, so a mismatch shows the stale notice and hides the form.

One thing that came out of it: aborting while loading with no earlier quote left the status on LOADING, so the button would sit on "Checking…" forever. That branch resets to IDLE now.

Tests: an interrupted request resolves as aborted and doesn't set SUCCESS, an interrupted re-quote keeps the earlier quote stale, and a quote for a different stay doesn't show the request form.

elcoleta (approved):

PR #3190, opened 2026-09-08
#3038 Fix blocked dates not being visible on direct booking websites

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3038

## Proposed Changes
Description:

Blocked dates on a published direct booking website looked the same as available ones. You couldn't click them, but nothing showed why, so a guest would try and get no response.

It's a regression from #3172. That PR made the calendar selectable, and .panoramaCalendarCellSelectable sets background: transparent. Same specificity as the reserved rule and later in the file, so it won. Every reserved cell got painted grey and immediately repainted transparent.

Two fixes:

The reserved styling now uses two class names, so it beats the single-class selectable rule regardless of where it sits in the file.

Reserved cells render as disabled with cursor: not-allowed instead of as enabled buttons with a pointer. They stay buttons with their "Reserved" aria-label, so screen readers still hear why they can't be picked, but sighted users now see it too. Before this the aria-label was the only place that said so.

Checked on a published site with the booking panel on: reserved dates are grey, not clickable, and neighbouring dates still work.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
Nothing refactored.

## Evidence
Reserved dates now show grey with a not-allowed cursor. Happy to add a screenshot if useful.

## Responsiveness
No layout change, only the fill and the cursor.

- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages

- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included — a reserved date renders as a disabled cell, clicking it doesn't start a stay, and the neighbouring day still works
  - [x] Jest tests are passing — 8 suites, 66 tests

## Keep or delete my branch
- [ ] Keep my branch


elcoleta (approved):
You can keep the visual reserved-state fix without disabling reserved dates unconditionally. The existing booking-selection logic allows a blocked/reserved date to be used as a valid checkout date when all preceding nights are available, since checkout is exclusive. With disabled: selectionState.isPast || selectionState.isReserved, the button becomes completely non-interactive before the selection logic can determine whether that reserved date is a valid checkout. This could introduce a regression where guests can no longer select an otherwise valid checkout date. Could we disable reserved dates only when they are not eligible to be used as checkout, while still preventing them from being selected as check-in dates?

karimh01:
Good catch, you were right. A reserved date can be a valid checkout when the nights before it are free, and my change blocked that.

Fixed it: a reserved date is now only disabled when it can't be a checkout. I used the same condition selectStayDate uses, so the button and the click logic can't disagree again.

Added a test for it too. With two dates blocked, both are disabled at first. Pick the day before and the first blocked date becomes clickable as checkout, the second stays disabled since there's a blocked night in between. Still grey either way.

EnesVeli (approved):
.panoramaCalendarCellSelectable.panoramaCalendarCellReserved at 0,0,2,0 beats the single-class rule regardless of file order, and confirmed the legacy calendar never had this bug so nothing's missed.

Also checked the isDisabled/canServeAsCheckOut logic against selectStayDate in quoteSelection.js — good catch keeping the checkout-exemption rule consistent between the two places it's now enforced. Mutation-tested the two new tests by reverting isDisabled to the old isPast-only check: both failed immediately, including the subtler day-2-enabled/day-3-disabled case, so this isn't just coverage padding.

Approving.

PR #3195, opened 2026-09-08
test(channex): add ARI push wire tests for channexProviderClient (#440)

EnesVeli:
## Close or Relate the Issue
- Closes #

- Related Issue: #440

## Proposed Changes
Description:
Slice 2 of 3 for ChannexProviderClient (functions/.shared/channelManagement/providers/channex/providerClient.js) — the HTTP boundary to the Channex API, which had zero unit coverage before this series because every consumer test mocks the class away.

This slice covers the ARI push path: pushAvailability, pushRestrictions, and the shared pushGroupedPayloads / postChannexPushRequest helpers they both delegate to. This is the most behaviourally complex code path in the file — it's the only one with batching, stopOnFailure, and AbortController timeout handling.

Covered:

Batch semantics — one request per group, one result per group; a group with empty values is rejected locally without a request while the rest of the batch still proceeds.
Both stopOnFailure break paths — the non-2xx response case and the thrown-request case, which are two separate branches in the loop (providerClient.js:310 and :326) and needed separate tests.
Status mapping unique to the push path — 401→UNAUTHORIZED and 429→RATE_LIMITED (the discovery-read methods only special-case 401), plus the method-specific fallback statuses.
Timeout handling — an aborted request maps to CHANNEX_PUSH_REQUEST_TIMEOUT, and no abort signal is attached at all when no requestTimeoutMs is configured.
Group identity — pushRestrictions carries externalRatePlanId, pushAvailability deliberately doesn't.

Three behaviours are pinned as characterization tests rather than "fixed", since changing them is a production decision and this is a tests-only PR:

A 2xx response carrying meta.warnings still reports success: false (success derives from warnings.length === 0, not the HTTP status).
An empty-values group never halts a batch even with stopOnFailure: true — that branch continues unconditionally; only real failures break.
An empty, undefined, or non-array batch returns { success: true, results: [] } — [].every(...) is vacuously true, so a caller reading only success can't distinguish "everything synced" from "nothing was sent". Worth knowing about given this path feeds certification demos.

Verified beyond a green run: I mutation-tested the three hardest assertions by temporarily breaking the production code — the abort→timeout mapping, and each stopOnFailure break independently — and confirmed the corresponding test (and only that test) fails each time, then restored. Went through the eng-review gate; both findings it raised are applied in this PR.

Slice 1 (discovery reads) is a separate open PR; Slice 3 (booking revisions) is next. No production code changed here.

## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size

## Refactoring
- src/features/example.xx

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x ] Conventions
- [x ] Testing
  - [x] Code tested locally
  - [x ] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch

### DIFF
--- a/backend/functions/UnifiedMessaging/business/channexProviderClient.ariPush.test.js
+++ b/backend/functions/UnifiedMessaging/business/channexProviderClient.ariPush.test.js
@@ -1,0 +1,382 @@
+const ChannexProviderClient = require("../.shared/channelManagement/providers/channex/providerClient.js").default;
+
+const originalFetch = global.fetch;
+const CREDENTIALS = { apiKey: "test-api-key" };
+
+const jsonResponse = (status, body) => ({
+  ok: status >= 200 && status < 300,
+  status,
+  text: async () => JSON.stringify(body),
+});
+
+const availabilityGroup = (overrides = {}) => ({
+  externalPropertyId: "ext-property-1",
+  externalRoomTypeId: "ext-room-1",
+  values: [{ date: "2026-06-01", availability: 1 }],
+  ...overrides,
+});
+
+const restrictionGroup = (overrides = {}) => ({
+  externalPropertyId: "ext-property-1",
+  externalRoomTypeId: "ext-room-1",
+  externalRatePlanId: "ext-rate-1",
+  values: [{ date: "2026-06-01", rate: "123.00" }],
+  ...overrides,
+});
+
+const PUSH_METHODS = [
+  {
+    method: "pushAvailability",
+    call: (client, groups, options) => client.pushAvailability(CREDENTIALS, groups, options),
+    callWithoutKey: (client, groups) => client.pushAvailability({}, groups),
+    group: availabilityGroup,
+    endpoint: "/api/v1/availability",
+    url: "https://staging.channex.io/api/v1/availability",
+    fallbackStatus: "AVAILABILITY_PUSH_FAILED",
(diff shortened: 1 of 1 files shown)


Mehedihasanb (approved):
Two small non-blocking suggestions:

1.The request tests assert the full https://staging.channex.io/... URL, while CHANNEX_BASE_URL is environment-configurable. It may be more robust to assert the pathname separately or explicitly control the base URL in the test environment.

2.The test named "a 401 thrown as an exception does not get the UNAUTHORIZED mapping" actually rejects with new Error("unauthorized") and does not contain an HTTP 401 response. Renaming it to something like "a rejected request does not get the HTTP UNAUTHORIZED mapping" would make the intent clearer.

karimh01 (approved):
Looks good, the stopOnFailure and timeout cases are the ones that matter and they're covered.

Comments in the file, same as the other PRs.

The one at line 157: a 2xx with warnings comes back as success:false. You say you're pinning it so it isn't mistaken for a bug, but is it not one? A successful push with warnings sounds like it should be success:true with warnings attached. Worth checking what the caller does with it before locking it in.

PR #3197, opened 2026-09-08
test(channex): add booking revision wire tests for channexProviderClient

EnesVeli:
…ent (#440, #2266)

## Close or Relate the Issue
- Closes #

- Related Issue: #440 #2266

## Proposed Changes
Description:

Slice 3 of 3 for ChannexProviderClient (functions/.shared/channelManagement/providers/channex/providerClient.js) — the HTTP boundary to the Channex API, which had no unit coverage before this series because every consumer test mocks the class away.

This slice covers the inbound booking-revision path: listBookingRevisionFeed, getBookingRevision, and acknowledgeBookingRevision, plus the shared normalizeChannexBookingRevision mapper all three depend on.

Covered:

Request construction — the feed's two query parameters (filter[property_id] and order[inserted_at]=asc), the single-revision path shape, the /ack suffix, and user-api-key headers throughout.
Endpoint mechanics — GET feed, GET by id, and POST ack. The ack is a POST that does not route through pushGroupedPayloads, and the test asserts init.body is undefined to prove it sends no payload.
Error mapping — 401→UNAUTHORIZED, generic non-2xx→method-specific fallback status, and network rejection fallbacks, as table-driven cases across all three methods. These three map 401 with an inline ternary rather than the getProviderStatusForHttpStatus helper the discovery reads use, so the branch needed its own coverage.
Response parsing — full normalized-shape mapping (ids, dates, amount, currency, rooms[], derived ratePlanId/roomTypeId, preserved rawPayload), id-less rows dropped, and the guestName fallback chain (customer.name → first+last joined → first only → null).

Two behaviours are pinned as characterization tests rather than "fixed", since changing them is a production decision:

A malformed feed body returns success: true with an empty revision list — not INVALID_RESPONSE, which is what the Slice 1 discovery-read methods do in the same situation. A silently empty feed reads very differently from an explicit error, and that asymmetry is now in writing.
getBookingRevision treats data: [] as INVALID_RESPONSE via an indirect route: typeof [] === "object" clears the guard, then the mapper finds no id on an array and falls into the empty-revision branch.

Verified beyond a green run: I mutation-tested both behaviours the plan called out as exact — made the malformed feed return a phantom row instead of degrading to [], and made the ack POST send a body — and confirmed each broke exactly its intended test and nothing else, then restored. Went through the eng-review gate.

Slices 1 and 2 are separate open PRs (discovery reads; ARI push). The fetch mock scaffolding is intentionally duplicated across the three files to keep the PRs fully decoupled — a shared test helper is planned as a follow-up refactor once all three merge. No production code changed here.
## Change Type
- [ ] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size

## Refactoring
- src/features/example.xx

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x ] Conventions
- [x ] Testing
  - [x ] Code tested locally
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch

### DIFF
--- a/backend/functions/UnifiedMessaging/business/channexProviderClient.bookingRevisions.test.js
+++ b/backend/functions/UnifiedMessaging/business/channexProviderClient.bookingRevisions.test.js
@@ -1,0 +1,340 @@
+const ChannexProviderClient = require("../.shared/channelManagement/providers/channex/providerClient.js").default;
+
+const originalFetch = global.fetch;
+const CREDENTIALS = { apiKey: "test-api-key" };
+const PROPERTY_ID = "ext-property-1";
+const REVISION_ID = "revision-1";
+
+const jsonResponse = (status, body) => ({
+  ok: status >= 200 && status < 300,
+  status,
+  text: async () => JSON.stringify(body),
+});
+
+const revisionRow = (overrides = {}) => ({
+  id: REVISION_ID,
+  attributes: {
+    booking_id: "booking-1",
+    property_id: PROPERTY_ID,
+    unique_id: "unique-1",
+    system_id: "system-1",
+    ota_reservation_code: "OTA-123",
+    ota_name: "Booking.com",
+    status: "new",
+    arrival_date: "2026-06-01",
+    departure_date: "2026-06-04",
+    arrival_hour: "15:00",
+    amount: "450.00",
+    currency: "EUR",
+    inserted_at: "2026-05-20T10:00:00Z",
+    customer: { name: "Ada Lovelace" },
+    rooms: [{ room_type_id: "ext-room-1", rate_plan_id: "ext-rate-1" }],
+    ...overrides,
+  },
+});
+
+const REVISION_METHODS = [
+  {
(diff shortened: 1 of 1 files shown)


Mehedihasanb (approved):
One small non-blocking suggestion: the feed request test currently asserts the full https://staging.channex.io/... URL. Since CHANNEX_BASE_URL is environment-configurable, this could make the test environment-dependent if that variable is set elsewhere. It may be more robust to assert the pathname and query parameters separately, or explicitly control the base URL in the test.

Also, the Refactoring section still contains the src/features/example.xx template placeholder and could be changed to N/A since this PR only adds tests.

BeMounir (approved):
looks good. only the issue bappy pointed out but its not a big issue. good job

PR #3205, opened 2026-09-09
Added my info to team page

PuoK24:
Added my name, role, and profile picture to the team page.


karimh01 (changes requested):
The build is already failing on this: Can't resolve '../../images/about-img/Puok.png'. Team.js imports Puok.png but the file you added is Puok.jpeg. Change the import to .jpeg, or rename the file.

Two other things that don't belong in this PR:

backend/CD/npm/createLambda.js has one space added to a blank line. That's a deploy script, nothing to do with the team page.

package-lock.json has 1727 lines added and 255 removed. Adding a photo doesn't need new packages, so it looks like npm install rewrote the whole lockfile, probably a different npm version than the rest of us. That can shift dependencies for everyone running npm ci. I'd put both files back:

  git checkout origin/acceptance -- frontend/web/package-lock.json
  git checkout origin/acceptance -- backend/CD/npm/createLambda.js

Then only the image and the two Team.js lines are left, which is what the PR is about.

Issue #2258, opened 2025-11-12
Host Dashboard Finance: Payouts: Invoice generation and details


stefanhopman1:
### Feature summary
---
These features will allow hosts and guests access to financial information using automatically generated invoices. These invoices will show general cashflow and deduction like channel commissions and management fees. 

---
### Categories

- [ ] #2900
- [ ] #2901 
- [ ] #2902 
- [ ] #2903 
- [ ] Later: #2904 
- [ ] Later: #2905 
- [ ] Later: #2906 
- [ ] Later: #2907 
- [ ] Later: #2908

PR #3210, opened 2026-09-10
feat(payments); add Stripe mode switch for test enviroments (only loc…

BeMounir:
## Close or Relate the Issue
- Closes #
- Related Issue: #2258

## Proposed Changes
Description:

Right now the backend reads the live Stripe key from SSM and the frontend has the live public key hardcoded. There's no way to run the payment flow against Stripe's sandbox without changing production.

This adds a switch. The backend reads /stripe/keys/secret/${STRIPE_MODE} instead of always /stripe/keys/secret/live. The frontend picks the public key based on REACT_APP_STRIPE_MODE. Without those env vars nothing changes, both default to live.

Also fixed while doing this: the refund path in reservationController built its own Stripe instance from STRIPE_SECRET_KEY, falling back to the test key from SSM. So refunds could try to refund a live PaymentIntent with the test key. Now refunds use the same client that created the payment.

**Nothing changes in production.** No env vars are set, so both sides stay on live.

**What this enables:** a second Lambda with STRIPE_MODE=test and TEST=true, pointing at the test schema, and a local frontend started with REACT_APP_STRIPE_MODE=SANDBOX. That's console work after this merges.

**Careful:** REACT_APP_STRIPE_MODE must never be set on the Amplify build. That would put production on sandbox.

## Change Type
- [x] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
The Stripe instance in reservationController is gone. It now asks bookingService for the shared client instead of building its own.

## Evidence
No UI change. Verified locally that without env vars the app still uses the live key.

## Responsiveness
Not applicable.

- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages

- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [ ] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [x] Keep my branch

### DIFF
--- a/backend/functions/General-Bookings-CRUD-Bookings-develop/business/bookingService.js
+++ b/backend/functions/General-Bookings-CRUD-Bookings-develop/business/bookingService.js
@@ -173,6 +173,10 @@ class BookingService {
     return parseBookingDateToMs(value, fieldName);
   }
 
+  getStripeClient() {
+    return this.stripeRepository.getClient();
+  }
+
   async confirmPayment(paymentid) {
     const booking = await this.reservationRepository.getBookingByPaymentId(paymentid);
     if (booking.status === BOOKING_STATUS_PAID) {
--- a/backend/functions/General-Bookings-CRUD-Bookings-develop/controller/reservationController.js
+++ b/backend/functions/General-Bookings-CRUD-Bookings-develop/controller/reservationController.js
@@ -1,7 +1,6 @@
 import BookingService from "../business/bookingService.js";
 import PaymentService from "../business/paymentService.js";
 import { BadRequestException } from "../util/exception/badRequestException.js";
-import Stripe from "stripe";
 import SystemManagerRepository from "../data/systemManagerRepository.js";
 import { calculateRefundAmountCents } from "../util/refundCalculator.js";
 import Forbidden from "../util/exception/Forbidden.js";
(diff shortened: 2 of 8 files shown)


karimh01 (changes requested):
Two things in stripeRepository.js that need fixing, both around line 22-28.

The fake Stripe object is broken. The backtick on `id` doesn't close until after testsecret, so the whole block ends up as one string assigned to id and there's no client_secret property anymore. Should be:

  id: `test_${randomUUID()}`,
  client_secret: `test_secret_${randomUUID()}`,

And line 22 says process.env.STRIPEMODE without the underscore. So useFakeStripe is always true when TEST=true, even with STRIPE_MODE=test set. On the test Lambda you'd get the fake instead of the real test client, which is the thing the switch was supposed to fix.

All the underscores in that spot are gone (test_, test_secret_, STRIPE_MODE), so it looks like the code went through something that reads underscores as italics.

Rest looks good. The frontend switch is right, getClient through the service is clean, and having the refund path use the same client now is a nice catch.

elcoleta (approved):
There looks to be an issue in the new Stripe-mode selection here. useFakeStripe checks process.env.STRIPEMODE, but the newly introduced variable is STRIPE_MODE. With the intended setup (TEST=true + STRIPE_MODE=test), STRIPEMODE is still undefined, so this condition keeps selecting the fake Stripe client instead of loading /stripe/keys/secret/test.
Also, in the fake paymentIntents.create() response, client_secret appears to have been moved inside the id template string rather than remaining a separate property. That means paymentIntent.client_secret will be undefined when createPaymentIntent() reads it.
Could we fix both in this block so the new test-mode flow actually uses the configured Stripe test client while preserving the existing fake client for unit tests?

karimh01 (approved):
Both fixed, checked it. The fake object has proper properties again and STRIPE_MODE has its underscore, so the fake only kicks in when no mode is set.

Approving.

Mehedihasanb (approved):
The existing StripeRepository has a TEST === "true" branch that returns a mocked Stripe client instead of loading a key from SSM. The PR description says the test Lambda would use both STRIPE_MODE=test and TEST=true. If that branch is still active, wouldn't TEST=true bypass /stripe/keys/secret/test completely?

It also looks like that mocked client only implements paymentIntents.create, so now that refunds use the shared client, could refunds fail in that TEST environment if they need the Stripe refunds API?

Small configuration suggestion as well: backend uses STRIPE_MODE=test while frontend uses REACT_APP_STRIPE_MODE=SANDBOX. Using the same live/test values on both sides and rejecting invalid explicit values would reduce the risk of accidentally falling back to live because of a typo.

Issue #3106, opened 2026-06-11
Direct Booking Website: Custom Domain (URL)


stefanhopman1:
## 🚀 Feature Summary

Allow hosts and property managers to connect and manage their own **custom domain names** for Domits Direct Booking Websites.

This enables businesses to operate under their own brand identity while leveraging the Domits website builder, booking engine, PMS, channel manager and payment infrastructure.

Examples:

- https://wellnessbisous.com/
- villatoscana.com
- bookings.luxuryvillasibiza.com
- stay.mykonoscollection.com

---

## 🧩 Problem Statement

Many property managers and luxury vacation rental brands want guests to book directly through their own branded website rather than a generic platform domain.

Without custom domain support:

- Brand trust is reduced
- Marketing effectiveness decreases
- SEO authority is fragmented
- Direct booking conversion rates may suffer
- Businesses cannot fully own their online presence

Luxury hospitality brands expect a professional and fully branded web experience.

---

## ✅ Desired Outcome

Hosts can connect their own domain name to their Domits direct booking website and operate under a fully branded URL while maintaining access to all Domits functionality.

---

## 👥 Target Users

- [x] Hosts / Property Managers / Enterprise
- [x] Marketing Teams
- [x] Revenue Managers
- [ ] Guests / Customers
- [ ] Developers

---

# 🚀 Key Features

---

## 1️⃣ Custom Domain Connection

Allow users to connect:

### Root Domains

Examples:

- villatoscana.com
- luxurystays.com

### Subdomains

Examples:

- book.villatoscana.com
- stay.luxurystays.com
- reservations.luxurycollection.com

---

## 2️⃣ DNS Configuration Wizard

Provide step-by-step setup guidance:

Supported records:

- A Record
- AAAA Record
- CNAME Record
- TXT Record (verification)

Features:

- DNS validation
- Domain verification
- Configuration status monitoring
- Troubleshooting guidance

---

## 3️⃣ SSL Certificate Management

Automatically provision and renew SSL certificates.

Features:

- HTTPS by default
- Automatic certificate issuance
- Automatic renewals
- Security monitoring

Result:

https://www.villatoscana.com

---

## 4️⃣ Multi-Domain Support

Support multiple domains per account.

Examples:

- villatoscana.com
- villasintuscany.com
- luxurytuscanystays.com

Use cases:

- Multiple brands
- Regional websites
- Marketing campaigns

---

## 5️⃣ Domain Redirect Management

Configure:

### 301 Redirects

[...]

PR #3213, opened 2026-09-10
#3106 Add CloudFront tenant repository for custom domains

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3106

## Proposed Changes
Description:

First backend part for custom domains. Three small things that the domain service in the next PR needs.

The CloudFront SDK, plus the lockfile changes that come with it. Four packages got removed because the newer SDK doesn't need them, and twenty others got a version bump. All backend tests still pass.

A repository that turns six CloudFront API calls into simple objects. The important bit: when it asks for a certificate, it always sets ValidationTokenHost to cloudfront. I tested this by hand today and the default setting, self-hosted, never works for a domain that has no other server. The tenant just gets stuck.

An error class built the same way as WebsiteQuoteError, so the controller can turn error codes into HTTP statuses later.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
Nothing refactored. The lockfile diff is large but it's npm reconciling dependencies, not code.

## Evidence
Verified a real GetDistributionTenant call works with the new SDK against developers.domits.com.

## Responsiveness
Not applicable.

## Npm Packages
- [x] NPM Packages installed
- [ ] NPM Packages removed
- [x] NPM Packages updated
- [x] Checked for vulnerabilities using "npm audit"

Added @aws-sdk/client-cloudfront. Lockfile bumped 20 transitive @aws-sdk and @smithy packages and dropped 4 that nothing needs anymore.

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included, 4 for the repository
  - [x] Jest tests are passing, 103 suites

## Keep or delete my branch
- [x] Keep my branch

Keeping the branch because the next PR stacks on it.

### DIFF
--- a/backend/functions/PropertyHandler/data/repository/cloudFrontTenantRepository.js
+++ b/backend/functions/PropertyHandler/data/repository/cloudFrontTenantRepository.js
@@ -1,0 +1,124 @@
+import {
+  CloudFrontClient,
+  CreateDistributionTenantCommand,
+  GetDistributionTenantByDomainCommand,
+  GetDistributionTenantCommand,
+  GetManagedCertificateDetailsCommand,
+  UpdateDistributionTenantCommand,
+  VerifyDnsConfigurationCommand,
+} from "@aws-sdk/client-cloudfront";
+
+const CLOUDFRONT_REGION = "us-east-1";
+const MANAGED_CERTIFICATE_VALIDATION_TOKEN_HOST = "cloudfront";
+const DNS_STATUS_UNKNOWN = "unknown-configuration";
+
+const isEntityNotFound = (error) => error?.name === "EntityNotFound";
+
+const mapTenant = (response) => {
+  const tenant = response?.DistributionTenant;
+  if (!tenant?.Id) {
+    return null;
+  }
+
+  return {
+    etag: response?.ETag || "",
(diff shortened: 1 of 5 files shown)


EnesVeli (approved):
Checked this against the actual SDK types (pulled @aws-sdk/client-cloudfront@3.1129.0) EntityNotFound and the response field mappings are all correct, and WebsiteCustomDomainError matches WebsiteQuoteError's shape properly. Nice fix hardcoding ValidationTokenHost: cloudfront.

applyCertificate sends UpdateDistributionTenantCommand with only Id/IfMatch/Customizations.Certificate. The type shape (everything but Id/IfMatch optional) suggests this is a real partial-update API rather than the old full-replace UpdateDistribution pattern, but the evidence section only mentions a real GetDistributionTenant call — did you get to verify applyCertificate against a real tenant too, to make sure it doesn't reset Domains/ConnectionGroupId/Enabled?

Everything else looks solid. Approving.

elcoleta (approved):

PR #3214, opened 2026-09-10
#3106 Add custom domain service for CloudFront tenants

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3106

## Proposed Changes
Description:

**Update:** the endpoints from #3231 were merged into this branch, so this PR now covers the service and the endpoints together. The endpoints were already reviewed and approved there.

The service that links a host's own domain to a CloudFront tenant and keeps track of the status. Builds on the repository PR.

**What I found out testing by hand:** getting a domain live takes three CloudFront calls, not one. Create the tenant with a certificate request, wait until the certificate is issued, then attach the certificate to the tenant with a second update. In the console that last step is hidden behind the Submit button. Through the API you have to do it yourself, and if you skip it the domain stays inactive forever. That's what this service handles.

Request: checks the hostname, refuses a domain that belongs to another site or a second domain for the same site, creates the tenant, and saves a CUSTOM row as PENDING with the CNAME the host has to set.

Sync: reads the tenant and certificate from CloudFront, attaches the certificate if it's issued but not on the tenant yet, and maps the state to PENDING, VERIFIED, ACTIVE, FAILED or DISABLED. Runs when the dashboard loads or the host clicks check. No scheduler yet.

Also two repository methods and a docs section that explains the whole flow, because that three-step thing isn't in AWS's docs and the next person would run into it too.

**Nothing changes in production.** Nothing calls this service yet, that's the endpoints PR.

**Console work before it can be used:**
- Three env vars on PropertyHandler: DIRECT_BOOKING_WEBSITE_CLOUDFRONT_DISTRIBUTION_ID, ..._CONNECTION_GROUP_ID, ..._ROUTING_ENDPOINT. Values are in the docs section.
- Six cloudfront actions on the Lambda role, listed in the docs.

**Known gap:** the rule "one custom domain per site" is only checked in the service, not in the database. A partial unique index on (site_id) WHERE domain_type = 'CUSTOM' would make it a real guarantee. That's a migration for later.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [x] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
Nothing refactored.

## Evidence
27 service tests covering every status transition, both error paths on tenant creation, and the throttling case.

## Responsiveness
Not applicable.

## Npm Packages

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included, 27 for the service
  - [x] Jest tests are passing, 103 suites

## Keep or delete my branch
- [ ] Keep my branch


EnesVeli (approved):
One thing worth calling out before merge: the description says "Nothing calls this service yet, that's the endpoints PR," but #3231 already merged into this branch, so the current diff actually does wire live endpoints — POST /property/website/domains, GET /property/website/domains, POST /property/website/domains/verify in index.js/propertyController.js, with host auth and site-ownership checks. That's real authenticated API surface, not inert service code, and worth a description update so reviewers know that's in scope here rather than assuming it's still to come. The auth/ownership handling in handleWebsiteDomainRequest looks right to me (401/403 mapped correctly, 404 on a site owned by someone else, error codes mapped through WebsiteCustomDomainError), but wanted to flag it explicitly since "nothing calls it yet" undersells what's actually being reviewed.

Approving on the code — just update the description to reflect that the endpoints are already in here.

BeMounir (approved):

Mehedihasanb (approved):
+1 on Enes's point — the description should say the host endpoints are already live in this diff (via #3231), not "nothing calls this service yet." Worth updating before merge so anyone reading this later understands the actual scope.

One thing I want to check on the service itself: in `createOrAdoptTenant`, the `EntityAlreadyExists` recovery path (tenant name `dbw-<siteId>` already taken) calls `getTenantByDomain(domain)` to adopt the existing tenant. That's recovering from a name collision with a domain lookup. It works for the documented case — retrying the exact same request after a failed DB write — but if the CloudFront tenant gets created, the DB write fails, and the host then retries with a *different* domain (say, a typo fix), the orphaned tenant is registered under the old domain, `getTenantByDomain(newDomain)` finds nothing, and the request dead-ends in `TENANT_CREATE_FAILED` with an orphaned tenant left behind. Is there a `getTenantByName` (or equivalent) on `CloudFrontTenantRepository` that would close this? If not, seems like the same category as the DB-constraint gap you already flagged — fine to ship with a follow-up noted, just want to make sure it's a known tradeoff rather than an oversight.

Btw this PR is 2700+ line.  Try to make smaller next time as the instruction.

Issue #2390, opened 2025-12-31
Property Manager - User Roles & Permissions - Customer IAM - AWS Cognito


stefanhopman1:
## 🚀 Feature Summary

Implement a scalable user roles and permissions system for **property managers managing 10–100+ properties**.

The system should support multiple teams, operational functions, and property-level access while remaining significantly simpler than the Enterprise authorization model.

The core model should separate:

1. **User**
2. **Role**
3. **Permission**
4. **Property / Portfolio Access**

The initial system should provide predefined roles while allowing users to belong to multiple roles and access multiple properties.

**Team**
- [x] #513
- [ ] #3239

---

## Problem Statement

Property managers operating 10–100+ properties typically have multiple employees and specialized teams.

A single Owner / Team Member permission model is no longer sufficient.

Different employees may be responsible for:

* Guest communication
* Reservations
* Revenue management
* Finance
* Housekeeping
* Maintenance
* Property operations
* Portfolio management

These users should not automatically receive access to every property or every part of the platform.

For example:

* A cleaner needs access to assigned properties and tasks.
* A guest-experience employee needs reservations and messaging.
* A revenue manager needs pricing and availability.
* A finance employee needs financial information.
* A manager may need access to an entire portfolio.
* An owner or administrator may need access to everything.

The system therefore needs scalable role-based and property-based authorization without introducing the full complexity of Enterprise RBAC.

> Focus on the problem, not the solution.

---

## Desired Outcome

A property management company should be able to manage **10–100+ properties and multiple employees** without creating security or operational complexity.

Success means:

* Users have clearly defined roles.
* Users can have access to specific properties or portfolios.
* Different departments can operate independently.
* Sensitive financial and administrative data is restricted.
* Managers can oversee multiple properties.
* Operational employees can work across assigned properties.
* Service providers can be limited to specific tasks/properties.
* Users can have multiple roles where required.
* Permission checks are enforced consistently across the application.
* The model can scale toward Enterprise without requiring a complete rewrite.

---

## Target Users

Who benefits from this feature?

* [ ] Guests / Customers
* [x] Hosts / Property Managers / Enterprise
* [x] Admin / Internal users
* [x] Developers
* [x] Other: Finance, Revenue, Operations, Housekeeping, Maintenance, Guest Experience teams

---

## Business & User Impact

How valuable is this feature?

* **Impact level:**

  * [ ] Low (nice-to-have)
  * [ ] Medium (improves workflows)
[...]

PR #3224, opened 2026-09-11
fix: validate host-team invite role against allow-list, excluding Hos…

Mehedihasanb:
## Close or Relate the Issue
- Related Issue: #2390

## Proposed Changes
Description:
Added server-side validation to `host-team`'s `inviteMember` so the `role` field must match a known allow-list before being persisted. Previously any string was accepted with no validation, and that value flows directly into the invitee's Cognito `custom:group` attribute when they accept the invite. This meant a direct API call with `role: "Host"` (bypassing the UI, which only ever sends `"Property Operations Manager"`) could grant a stranger Host-group privileges through another user's invite.

Changes:
- New `backend/functions/host-team/util/roles.js` — frozen `ALLOWED_TEAM_MEMBER_ROLES` allow-list (8 roles) and `DEFAULT_TEAM_MEMBER_ROLE`. `Host`, `Admin`, and `Traveler` are intentionally excluded.
- `backend/functions/host-team/business/service/service.js` — added `normalizeRole()`, following the same pattern as `communication-preferences`'s `normalizePersona`: defaults to `Property Operations Manager` when falsy, throws `BadRequestException` when the value isn't in the allow-list. Wired into `inviteMember`. Changed the constructor to `constructor({ repository = new Repository() } = {})` for dependency injection (needed for the new tests; the only existing caller, `controller.js`, is unaffected since the param is optional).
- New `backend/test/host-team/service.test.js` — 13 tests covering `normalizeRole` (defaults, all valid roles accepted, `Host`/`Admin`/`Traveler`/unknown strings rejected) and `Service.inviteMember` end-to-end for the same cases, asserting `repository.create` is never called on rejection.

This is a backend-only fix and intentionally scoped separately from the frontend invite-dropdown work (also under #2390), which is pending a product decision on whether "Host" should ever be invitable and will be a separate PR.

## Change Type
- [x] Bug fix

## Change Size

## Refactoring
N/A — no files changed without logic changes.

## Evidence

N/A — backend-only change, no UI affected.

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

N/A — backend-only change.

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

N/A — no package changes.

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch

### DIFF
--- a/backend/functions/host-team/business/service/service.js
+++ b/backend/functions/host-team/business/service/service.js
@@ -3,20 +3,31 @@ import { BadRequestException } from "../../util/exception/badRequestException.js
 import { ForbiddenException } from "../../util/exception/forbiddenException.js";
 import { NotFoundException } from "../../util/exception/notFoundException.js";
 import { sendTeamInviteEmail } from "../emailService.js";
+import { ALLOWED_TEAM_MEMBER_ROLES, DEFAULT_TEAM_MEMBER_ROLE } from "../../util/roles.js";
 import { CognitoIdentityProviderClient, AdminUpdateUserAttributesCommand, AdminGetUserCommand } from "@aws-sdk/client-cognito-identity-provider";
 import Database from "database";
 
 const cognitoClient = new CognitoIdentityProviderClient({ region: "eu-north-1" });
 const USER_POOL_ID = "eu-north-1_mPxNhvSFX";
 
+export const normalizeRole = (role) => {
+    if (!role) return DEFAULT_TEAM_MEMBER_ROLE;
+    const normalized = String(role).trim();
+    if (!ALLOWED_TEAM_MEMBER_ROLES.includes(normalized)) {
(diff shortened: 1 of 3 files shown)


Nishchal-code (approved):
Looks clean

karimh01 (approved):
Good catch, this is a real one. A host could invite someone as Admin and acceptInvite would write that straight into Cognito. Allow-list is the right fix, and I checked, inviteMember is the only path that sets a role so it covers everything.

One thing: the four-line comment in roles.js. No comments in code. But that explanation matters, so put it in the commit message where the why of a security fix belongs.

PR #3226, opened 2026-09-11
feat: restructure host-team member list into a responsive table

Mehedihasanb:
## Close or Relate the Issue
- Related Issue: #2390

## Proposed Changes
Description:
Restructured the host-team member list from a flat row (avatar + email + role badge + remove button) into a full table: Co-Host (avatar/name/email), Role (colored badge), Permissions (display-only summary), Status (Active/Pending), and an actions menu. Previously, active and pending members were split into two separate sections with no dedicated status indicator — status is now shown per-row as a column instead.

The "Permissions" column is cosmetic labeling only, sourced from a new static lookup (`ROLE_PERMISSIONS_SUMMARY`) — it does not affect or introduce any enforcement. Real authorization remains the existing ownership/role checks in the backend auth layer, unchanged by this PR. The role list feeding this table (`TEAM_INVITABLE_ROLES`) is derived from `HOST_ROLES` filtered against Admin/Host, matching the same exclusions already enforced server-side in #2390's backend allow-list PR.

Locale strings added across en/nl/de/es for the new column headers, loading/empty/error states, and actions menu; the now-unused `activeMembers`/`pendingInvitations` keys were removed from all four locale files.

Invite modal and role dropdown are intentionally untouched — that remains a separate pending task.

## Change Type
- [x] New feature

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
N/A — no files changed without corresponding logic/markup changes.

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

N/A — no package.json changes.

## Checklist
- [ ] Conventions
  - [ ] No global styling (see [#1691](https://github.com/domits1/Domits/issues/1691))
- [ ] Testing
  - [ ] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


karimh01 (changes requested):
The table restructure looks good and the tests are thorough.

But the role editing can't work: updateTeamMemberRole does a PUT to the team API, and the backend has no PUT handler. The service has inviteMember, removeMember, getTeamMembers, getMemberships and acceptInvite. Nothing that updates a role. So the button either does nothing or errors.

When that endpoint gets built, it has to go through the same allow-list #3224 just added to inviteMember. Otherwise a host can invite someone as a manager and then edit them to Admin. The frontend filters the dropdown, but that's not protection.

Either drop the role editing from this PR and ship the table, or add the backend endpoint with the allow-list check. As it stands it ships a button that doesn't work.

Also: two comment blocks in rolePermissionsSummary.js. No comments in code.

Mehedihasanb:
Backend endpoint is in — `updateMemberRole` now exists in service.js/controller.js/index.js, 
routed via PUT, going through the same allow-list as `inviteMember` (#3224), plus a 
caller-authorization check restricting it to Host/Admin. Tests added covering the allow-list, 
missing-role rejection, ownership check, and the caller-authorization case specifically 
(non-Host/Admin caller rejected even with a fully valid role). See 43ac2b6.

On the comments in `rolePermissionsSummary.js` — checked comment density across `hostdashboard/` 
first: only 5 of 201 files have any comments at all, so the density point is fair. But those 
5 are short "why" notes, same style as these two, matching CLAUDE.md's own convention 
("comment why, not what"). Both blocks here explain something that isn't recoverable from 
the code alone — why Admin/Host are excluded (tied to the Cognito write in acceptInvite) 
and that these permissions are display-only, not enforced. I'd like to keep them unless 
there's a reason I'm missing — happy to trim if they feel too long.

Nishchal-code (approved):

karimh01 (approved):
Both my points from last week are fixed. PUT is routed (index.js:41-43) and both invite and edit go through normalizeRole against the allow-list (service.js:13-20, roles.js:9-18), which has no Host or Admin. The backend tests prove those are rejected and never persisted. Ownership holds everywhere: remove and role-update both refuse when member.host_id does not match, and the caller id comes from Cognito on the access token, never from the request.

Three things still open, and the first one sends wrong information to real people.

**The invite email always says Property Operations Manager.** You widened the role dropdown (HostTeam.js:458-463) but emailService.js:16,18 still writes that one role into the mail. So inviting a Financial Manager sends them an email saying they are a Property Operations Manager. Either make the email role-aware or narrow the dropdown back until it is.

**The table overflows its card between 641px and roughly 720px.** The grid minimum is 634px of tracks plus 48px of gaps (settingsDashboard.css:50), the card has 20px side padding (:545), and nothing sets overflow-x on .team-table or .team-card. The stacked layout only kicks in at max-width 640px, so there is a band just above the breakpoint where the row spills out. Either lower the grid minimum or add overflow-x: auto.

**A malformed PUT body answers 500.** JSON.parse at controller.js:75 throws a plain error and handleError (:84-90) has no status, so it becomes a 500. Same pattern already exists in inviteMember (:27), so it is not new, but worth mapping to 400 while you are in there.

Smaller: member_name is returned by the service (service.js:88-93) but the row only shows member_email (HostTeam.js:92). And Cancel and Escape stay live while a save is in flight (:208-213,512) while the completion handler closes whatever modal is open at that moment (:263), so a slow save for one member can close or mislabel another. Low frequency, but real.

On the Cognito sync Copilot raised: I would make that a follow-up issue rather than block on it. updateMemberRole writes the table only, but the only backend that reads the specific manager role (property-tasks pomService.js:5-6) treats every manager role as one set, so an edit changes the label and not what the member can do. Worth fixing before anything starts depending on the specific role.

Tests are better than I expected. The backend covers every allow-listed role persisting, rejection of Host, Admin and unknown roles, a missing member, another host's member, and a non-Host caller. Not covered: the edit-modal failure path, the remove failure, and malformed bodies.

On the comments: I am not going to keep pushing it. The team convention is no comments in code, and there are still blocks in rolePermissionsSummary.js:3-6,11-15, HostTeam.js:23-25 and util/roles.js:3-8. If you disagree with the convention, raise it with Robert rather than in this PR, because right now it is inconsistent with the rest of the repo.

PR #3228, opened 2026-09-11
fix(bookings): make acceptInquiry atomic and race-safe on Aurora DSQL 

### LINKED ISSUE #3211: [Bug]: bookingService.acceptInquiry commits AWAITING_PAYMENT before the overlap search runs (non-transactional race window)
## Bug Summary
`acceptInquiry` writes the accepted booking's status to `AwaitingPayment` *before* it looks for and declines overlapping inquiries on the same property/dates. The two writes are not wrapped in a transaction or lock, opening a race window between them.

- Parent Issue: #440 
- Related Issue: #2266 
---

## Expected Behavior
Accepting an inquiry and declining its overlapping siblings should be atomic: either both the status commit and the overlap-decline step succeed together, or neither does. Two overlapping inquiries should never both reach `AwaitingPayment`.

---

## Actual Behavior
Describe what actually happened.
In `backend/functions/General-Bookings-CRUD-Bookings-develop/business/bookingService.js`, `acceptInquiry(bookingId, authToken)`:

```js
269  async acceptInquiry(bookingId, authToken) {
...
277    await this.reservationRepository.updateBookingStatus(bookingId, BOOKING_STATUS_AWAITING_PAYMENT);
278
279    const overlapping = await this.reservationRepository.getOverlappingInquiries({
280      propertyId: booking.property_id,
281      arrivalDateMs: booking.arrivaldate,
282      departureDateMs: booking.departuredate,
283      excludeBookingId: bookingId,
284    });
285    await Promise.all(overlapping.map((b) => this.reservationRepository.updateBookingStatus(b.id, "Declined")));

Line 277 (status commit) runs before lines 279–285 (overlap search + decline), with no transaction spanning them and no compensating action if either later step fails. As a result:

A second, concurrent acceptInquiry call on an overlapping inquiry for the same property/dates can also pass its own status !== "Inquiry" check (both bookings still read "Inquiry" before either write lands), so both can be promoted to AwaitingPayment.
If the process throws between the status commit and the decline loop (e.g. getOverlappingInquiries fails, or one of the updateBookingStatus calls in the Promise.all rejects), the accepted booking is left in AwaitingPayment while the overlapping inquiries that should have been declined never are — leaving multiple live paths to payment for the same dates.
---

## Steps to Reproduce
Steps to reproduce the behavior:
1. Create two overlapping inquiries (Inquiry status) for the same property and overlapping date ranges.
2. Call POST accept-inquiry for both bookings at effectively the same time (or inject a delay/failure between lines 277 and 285 for one of the calls).
3. Observe both bookings' final status.
4. Both can end up in AwaitingPayment instead of exactly one, or one can be stuck in AwaitingPayment with its overlaps never declined if step 279–285 fails after step 277 commits.

---

## Environment
Select the environment where the issue occurs:
- [ ] Production
- [ ] Staging
- [x] Development

---

## Severity
Select one:
- [ ] Low (minor UI issue)
- [ ] Medium (feature degraded)
- [x ] High (core feature broken)
- [ ] Critical (system outage / data risk)

---

## Security & Data Considerations
Check all that apply:
- [ ] Authentication / authorization involved
- [x ] Personal or sensitive data involved
- [ ] Could be exploitable
- [ ] No security implications identified

---

## App / Service
Which part of the system is affected?  
Backend — General-Bookings-CRUD-Bookings-develop (bookingService)

Acceptance Criteria
[ ] The status transition and the overlap-decline step are made atomic (single transaction, or a re-check of the target booking's current status inside the same transaction/lock before committing AwaitingPayment).
[ ] A concurrent-acceptance scenario (two overlapping inquiries accepted at effectively the same time) results in exactly one booking reaching AwaitingPayment and the other(s) declined — never both.
[ ]  A failure during the overlap search or decline step does not leave the target booking committed to AwaitingPayment with un-declined overlaps; either the whole operation rolls back, or the failure is surfaced and retried consistently.
[ ]  A regression test covers: (a) the atomic success path, (b) a simulated failure between the status write and the overlap decline, and (c) two concurrent acceptInquiry calls on overlapping inquiries.

EnesVeli:
## Close or Relate the Issue
- Closes #3211 

- Related Issue: #440 

## Proposed Changes
Description:

`acceptInquiry` moved the booking to `Awaiting Payment` first and only then looked for overlapping inquiries to decline, in separate, non-transactional steps. Two overlapping inquiries accepted at the same time could both reach `Awaiting Payment` (a double booking), and a failure halfway through left the accepted booking committed with its overlaps never declined.

**What changed**
- **One atomic transaction.** New `reservationRepository.acceptInquiryWithOverlapDecline` does the whole accept in one `client.transaction()`: lock the conflict set, accept the target, decline every overlap. A failure at any point rolls back everything.
- **Locked read before any write.** `SELECT ... FOR UPDATE` (`setLock("pessimistic_write")`) reads the target plus every overlapping `Inquiry` booking on the property. If the target is no longer `Inquiry`, nothing is written and the host gets the existing "Booking is not in Inquiry status" 400.
- **Retry on DSQL conflicts.** Aurora DSQL uses optimistic concurrency, so nothing blocks. When two overlapping accepts run at once, the second commit fails with SQLSTATE `40001` (`OC000`) and rolls back entirely. The repository retries only on `40001` (up to 3 attempts). The retry re-reads committed state, finds its target already declined, and returns `accepted: false`, so the losing host gets a clean 400 instead of a raw 500.
- **Business layer** (`bookingService.acceptInquiry`) now calls the atomic method instead of the old three-step sequence. The authorization checks (not found / not the host / not Inquiry) are unchanged.
- **Removed dead code:** `getOverlappingInquiries` had no callers left once `acceptInquiry` was rewired.

**How the database behaviour was verified**
Checked by hand against the real DSQL cluster (`test` schema only) before writing the fix:
- `pg_advisory_xact_lock`: not supported
- `SERIALIZABLE` isolation: rejected (default is `repeatable read`)
- Exclusion constraints: not possible, since `CREATE EXTENSION` (`btree_gist`) is unsupported
- `SELECT ... FOR UPDATE`: supported, so it's the primitive used here

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Big change (+-max 1000)

+409 / −17 across 5 files. About 70 lines are production code; the rest are tests.

## Refactoring
No code was moved without changes. `getOverlappingInquiries` was deleted after `acceptInquiry` stopped using it (it had no other callers). Its overlap query now lives, with a lock added, inside the new transaction.

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

### Known follow-up (not in this PR)
The create path has a separate, narrower double-booking window. `acceptInquiry` only checks overlapping *Inquiry* bookings, so an inquiry created at the same moment an overlapping one is accepted can still be accepted later. Tracked in its own issue.

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


karimh01 (approved):
Good fix. DSQL doesn't do blocking locks, so this is the way: lock the rows, let the second one fail on commit, try again. On the retry the loser sees its inquiry is already declined and stops. That's exactly what you want.

Two things:

Five lines of comment above the function. No comments in code. It's useful stuff though, so put it in the commit message.

The declines happen one by one in a loop. One UPDATE with WHERE id IN would do the same. Fine with three overlaps, not with thirty.

One thing the title makes bigger than it is: this is safe between inquiries, but the lock only grabs rows with status Inquiry. A Confirmed booking on the same dates isn't in there. So a host can still accept an inquiry on top of a confirmed booking. That was already the case, so not your bug, but worth a follow-up.

BeMounir (approved):

PR #3235, opened 2026-09-14
Fix/3212 channex empty batch success

### LINKED ISSUE #3212: [Bug]: ChannexProviderClient ARI push (pushAvailability/pushRestrictions) reports success:true on empty batches
## Bug Summary
The shared push helper behind `pushAvailability`/`pushRestrictions` derives `success` with `Array.prototype.every`, which returns `true` on an empty array. When there is nothing to push, zero requests are sent to Channex but the result — and the evidence row derived from it — still reports success.
- Parent Issue: #440 
- Related Issue: #2266 
---

## Expected Behavior
An ARI sync that sends zero outbound groups (no dates/values assembled to push) should be reported as something other than a genuine success — either an explicit `NOOP`/empty-batch status, or `overallSuccess: false` with a clear reason — and should be distinguishable from a run that actually pushed data to Channex.
---

## Actual Behavior
`backend/functions/.shared/channelManagement/providers/channex/providerClient.js:331`, in `pushGroupedPayloads` (feeding `pushAvailability` at line 696 and `pushRestrictions` at line 716):
```js
return {
  success: results.every((result) => result.success),
  results,
};

[].every(...) is vacuously true. If groups is empty, the for loop never runs, results stays [], and this returns success: true with zero requests actually sent.

The same failure mode is duplicated one layer up in backend/functions/.shared/channelManagement/utils/channexAriExecutionUtils.js:239-260, deriveEvidenceOutcome. buildChannexSyncProviderResult (channexAriExecutionService.js:356-361) calls it without ever passing overallSuccess, so it's undefined inside. With an empty results array, failedCount, successfulCount, and warningCount are all 0, none of the earlier branches match, and it falls through to:

260  return { status: "SUCCESS", overallSuccess: true };

So a zero-group ARI sync is persisted to channexEvidence as status: "SUCCESS", indistinguishable from a run that genuinely pushed rate/availability changes.
---

## Steps to Reproduce
Steps to reproduce the behavior:
1. Trigger an ARI (availability or restrictions) sync where the upstream payload-grouping step produces zero groups (e.g. an empty date range, or a mapping/filter bug that drops every payload before it reaches the provider client).
2. Inspect the API response and the corresponding channexEvidence row for that sync.
3. Observe success: true / status: "SUCCESS" / requestCount: 0 — reported as a successful sync despite nothing being sent to Channex.
---

## Environment
Select the environment where the issue occurs:
- [ ] Production
- [x ] Staging
- [x ] Development

---

## Severity
Select one:
- [ ] Low (minor UI issue)
- [x ] Medium (feature degraded)
- [ ] High (core feature broken)
- [ ] Critical (system outage / data risk)

---

## Security & Data Considerations
Check all that apply:
- [ ] Authentication / authorization involved
- [ ] Personal or sensitive data involved
- [ ] Could be exploitable
- [x ] No security implications identified

---

## App / Service
Backend — Channex channel management (.shared/channelManagement, ARI push path)

[ ] pushGroupedPayloads (and therefore pushAvailability/pushRestrictions) reports success: false (or a distinct explicit status, e.g. NOOP/EMPTY_BATCH) when groups is empty, rather than defaulting to true via vacuous every().
[ ] deriveEvidenceOutcome receives an explicit overallSuccess from its caller and does not fall through to SUCCESS when results is empty and nothing was actually sent.
[ ] The evidence row (channexEvidence) for a zero-group ARI run is distinguishable from a genuinely successful push.
[ ] Regression tests cover: pushAvailability/pushRestrictions called with an empty groups array, and deriveEvidenceOutcome called with an empty results array and no overallSuccess passed.

EnesVeli:
## Close or Relate the Issue
- Closes #3212 

- Related Issue: #

## Proposed Changes
Description:
`ChannexProviderClient.pushAvailability`/`pushRestrictions` computed success as
`results.every((r) => r.success)`. `[].every(...)` is vacuously true in JS, so an
empty batch (no groups, or a non-array/undefined argument) reported `success: true`
with zero results — indistinguishable from "everything synced." The same class of
bug existed one layer up in `deriveEvidenceOutcome`, which fell through to the same
`SUCCESS` status when `calledProvider` was true but `results` was empty.
Both are fixed at the source instead of relying on every caller remembering its own
guard:
- `providerClient.js`: `success` now requires `results.length > 0` in addition to
  every result succeeding.
- `channexAriExecutionUtils.js`: `deriveEvidenceOutcome` now returns `NOOP` (the
  same status already used when the provider isn't called at all) when the
  provider was called but returned zero results.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
- src/features/example.xx

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


Nishchal-code (approved):

BeMounir (approved):
Looks good

PR #3240, opened 2026-09-14
fix(ical-sync): report misconfigured sources as skipped instead of succeeded

### LINKED ISSUE #3199: Ical-sync-scheduler: misconfigured sources silently count as succeeded
## Problem Statement

When the iCal sync job runs, it goes through every calendar source that's
due and tries to refresh it. If a source is missing something it needs — a
property ID, a source ID, or the calendar URL itself — the job quietly skips
it and moves on. The problem is that skip gets counted exactly the same as a
success. Open the sync summary afterward and there's no way to tell "this
calendar synced fine" apart from "this calendar was never even attempted
because someone's data is broken."

Practically, that means a source with a typo'd URL, or one that lost its
source ID in a bad migration somewhere, can sit there for weeks looking
perfectly healthy in every report we'd actually look at.

## Where This Happens

`backend/functions/Ical-sync-scheduler/business/service/service.js`

The function that refreshes a single source (`refreshSingleSource`) checks
for those three required fields up front, and if any is missing, it just
returns — no error, nothing thrown. That's reasonable on its own; we don't
want a bad source to blow up the whole sync run.

The problem is one level up. The code that calls `refreshSingleSource`
wraps it in a try/catch and only knows two outcomes: it either threw
(recorded as failed) or it didn't (recorded as succeeded). Since a skipped
source never throws, it falls into the same bucket as a real success. There's
currently no third bucket for "we didn't actually try this one."

## What We Want Instead

We want to be able to look at a sync run and see three outcomes, not two:
sources that actually synced, sources that tried and genuinely failed
(network error, bad response, that sort of thing — already handled), and
sources that were skipped because their setup is broken. Right now that
third case is invisible.

## Acceptance Criteria

- A source that's missing a property ID, source ID, or calendar URL is
  reported as **skipped**, not succeeded.
- The summary the sync job returns includes a skipped count, alongside the
  existing succeeded and failed counts.
- Sources that were actually attempted keep behaving exactly as they do
  today — this only changes what happens to the ones that never got that
  far.
- Add test coverage for the new skipped outcome. There's already a test in
  `backend/test/Ical-sync-scheduler/service.conflict.test.js` that covers
  this exact scenario today — it currently expects the source to count as
  succeeded, so it'll need updating once this changes.

## How This Came Up

A reviewer caught this while looking at a test-coverage PR (#3180, under
#2266). The test was accurately describing what the code does today, and the
reviewer asked, reasonably, whether that was on purpose. It isn't — it's just
a case nobody handled. Filing it separately here because fixing it means
changing production behavior, which didn't belong in a PR that was only ever
supposed to be about adding tests.

EnesVeli:
## Close or Relate the Issue
- Closes #3199 

- Related Issue: #

## Proposed Changes
Description:
`refreshSingleSource` silently returned (no throw, no signal) when a source was
missing its property ID, source ID, or calendar URL. The caller in
`runScheduledSync` only had two buckets — threw → failed, didn't throw →
succeeded — so a skipped source was indistinguishable from a real success. A
misconfigured source (typo'd URL, a source ID lost in a bad migration) could
sit looking perfectly healthy in every sync report indefinitely.

`refreshSingleSource` now returns an explicit outcome (`SYNC_OUTCOME.SYNCED` or
`SYNC_OUTCOME.SKIPPED`), and the worker adds a third bucket (`SYNC_OUTCOME.FAILED`
on a thrown error). The summary `runScheduledSync` returns now includes a
`skipped` count alongside `succeeded` and `failed`.

Sources that are actually attempted behave exactly as before, this only
changes what happens to the ones that never got that far.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
- src/features/example.xx

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


BeMounir (approved):

karimh01 (approved):
Good catch. A source with no URL or ID counted as a success because refreshSingleSource returned nothing and that read as ok. Now it's a separate outcome, so the report tells you what actually happened.

Three states in a frozen object instead of a boolean is the right shape for this, easier to add a fourth later.

Approving.

PR #3245, opened 2026-09-15
fix(tests): repair acceptance by dropping inquiry tests superseded by the atomic accept path

EnesVeli:
## Close or Relate the Issue
- Closes #

- Related Issue: #3238 

## Proposed Changes
Description:
`acceptance` is currently failing 10 backend tests across two files, blocking every
backend PR.

#3228 replaced `getOverlappingInquiries` with the transactional
`acceptInquiryWithOverlapDecline`. #3181 merged after it, carrying tests still written
against the old method. Both PRs were green on their own branches — backend tests don't
run on `acceptance` itself, so the semantic conflict only surfaced after both landed.

This is a trim rather than a rename, because most of the failing tests were superseded
rather than just renamed:

**`reservationRepository.overlappingInquiries.test.js` — deleted.** All five query
clauses it asserted are already asserted in `reservationRepository.acceptInquiry.test.js`
(lines 61–66), which additionally asserts the `pessimistic_write` lock the old file
couldn't. The one clause with no equivalent, `booking.id != :excludeBookingId`, is gone
by design: exclusion moved from SQL into JS, and the line-36 test proves the target gets
`Awaiting Payment` while only the others get `Declined`.

**`bookingService.inquiry.test.js` — trimmed to its `declineInquiry` coverage.** That
turned out to be the only place `declineInquiry` is tested anywhere in the suite, so it
stays. Its `acceptInquiry` cases duplicate `bookingService.acceptInquiry.test.js`, with
one exception worth calling out:

> `"leaves the accepted booking committed even when the overlap search fails"`

That test asserted the exact non-transactional behaviour #3211 fixed. Renaming it would
have re-enshrined the bug, so it's deleted — the corrected behaviour is asserted at
`reservationRepository.acceptInquiry.test.js:133` ("propagates a failure from declining
an overlap so the transaction rolls back the target's acceptance too").

The file also now sets `TEST=true`, matching its sibling. Without it the suite reached
real SSM and crashed the Jest worker with `ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING_FLAG`.

**Process note for the team:** backend tests not running on `acceptance` is what let two
independently-green PRs break the branch. Worth considering a scheduled run on
`acceptance` so the next semantic conflict is caught at merge rather than by whoever
pulls next.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
- src/features/example.xx

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


karimh01 (approved):
Checked it. The deleted tests were duplicates of acceptInquiry.test.js, and that one covers the lock too which the old one couldn't. The one you dropped from inquiry.test.js was testing the bug #3228 fixed, so renaming would have locked the old behaviour back in. Keeping declineInquiry there makes sense since nothing else covers it.

Only one test left in that file now. Fine, just thin.

Approving. And yes to tests on acceptance, that's the real fix.

elcoleta (approved):
Looks good. The removed acceptInquiry coverage is already represented by the newer atomic transaction tests, including the locking and rollback behavior, while the unique declineInquiry coverage is preserved. The TEST=true setup also makes sense for keeping this suite isolated from SSM. Approving

PR #3246, opened 2026-09-15
#3038 Warn the host when the listing changed after the last publish

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3038

## Proposed Changes
Description:

A host who changes their listing doesn't see it on their live site until they click "Update live site". Nothing told them that. I ran into it myself: replaced the photos on a listing, the live site showed broken images until I republished.

Now the website tab shows a notice on a published site when the listing has changed since the last publish: "Your listing changed after the last publish (date). Guests still see the older version. Open the editor and use Update live site." The same notice shows in the editor's Live site panel.

**How it's checked:** the site summary already carries a snapshot of the listing as it was at publish time, and the website tab already fetches the current listing per card for the thumbnail. The notice compares the two. I first planned to compare timestamps, but the property's updated_at only changes on title, description and status edits, not on photo changes. So that would have missed the exact case I hit.

**One thing fixed along the way:** the "Update live site" button in the editor was disabled unless the website's own overrides had changed. A host whose only change was the listing couldn't republish from there at all. It now enables when either changed.

**What it costs:** one extra site summary request per published site when the website tab loads. Fine for a handful of sites per host.

**Known limit:** the card checks when the tab loads. If a host changes photos in another tab and comes back without leaving the page, the card keeps its earlier answer until it refreshes. The editor recomputes on every open.

## Change Type
- [x] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
Nothing refactored.

## Evidence
Tested locally: changed the description of a published listing, went to the website tab, saw the notice. Clicked Update live site, notice gone.

## Responsiveness
One notice block with the existing card styling.

- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [x] UI checked on desktop

## Npm Packages

- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included, 15 in websiteListingChange.test.js
  - [x] Jest tests are passing, 192 in the website folder

## Keep or delete my branch
- [ ] Keep my branch


Mehedihasanb (changes requested):
Nice fix, and the test coverage on `websiteListingChange.js` is thorough — the false-positive guards (availability, host profile, whitespace-only edits, amenity reordering) are exactly the cases I'd worry about.

One thing I want to check before approving: buildListingDigest() runs on both `site.publishedPropertySnapshot` and the freshly-fetched `propertyDetails`, assuming they're shaped the same way. The tests only ever feed it one hand-built fixture for both sides, so they can't catch a shape mismatch. If `publishedPropertySnapshot` doesn't structurally match what `fetchWebsitePropertyDetails` returns, every already-published site could show "changed since publish" immediately on deploy, even with nothing changed.

Two smaller questions, not blockers:

- `IMAGE_KEY_FIELDS` in `websiteListingChange.js` prioritizes `key` over `web_key`/`url`/`src`. Is `key` actually what the live site template renders from? If the template uses `web_key` and that gets regenerated independently, staleness could go undetected — the same failure mode this PR is fixing.
- The pricing digest only compares `roomRate` and `cleaning`. Does the site show any other pricing fields (min stay, weekly rate, extra-guest fee)? If so those changes wouldn't trigger the notice either.

Nit: `pruneWebsiteDraftPreviewModels` gets reused to prune `websiteDraftLiveSiteStates` in `WebsiteBuilderPage.js` — works, but the name is now misleading at that call site.

elcoleta (changes requested):
The stale comparison seems to miss some listing fields that are actually rendered on the published site. buildWebsiteTemplateModel uses availabilityRestrictions (e.g. MinimumStay) and check-in/check-out values, but buildListingDigest() doesn’t include them.
This means changing something like the minimum stay could leave the live site on the old value without showing the stale notice, and Update live site may remain disabled when there are no editor override changes. Could we include the remaining listing fields that affect the published template in the digest?

Mehedihasanb (approved):
Approved . All three things I raised are handled correctly.

karimh01:
Thanks, checked all of it. Two of your three points were right and I fixed them.

**Shape mismatch between the two sides**

This one holds up. Both payloads come from the same function. The snapshot is built in propertyController.js:1920 with getFullPropertyAttributesWithFullLocation(propertyId, { includeCalendarAvailability: true }) and stored on line 1938. fetchWebsitePropertyDetails targets hostDashboard/single, which index.js:68 routes to getFullOwnedPropertyById, through getFullPropertyByIdAsHost, into the same function with the same options (propertyService.js:264-271). Both sides get JSON.stringify'd.

The model classes look off with this._roomRate, but those are setters that assign public fields and no model defines toJSON, so serializing gives roomRate, cleaning, amenityId, detail/value and rule/value. Location goes through getFullLocation into a PropertyLocation with public city and country. The frontend enrichment only spreads and adds, it renames nothing. So no notice storm on deploy.

Worth adding though: the digest is an allow-list, so new fields on the producer are ignored, but a rename is not the only thing that can break it silently. If the way an already-read field is derived changes it also flips, for example propertyImageRepository.js:101 sets key to web?.key || original?.key. I diffed the mapping files, models and repositories since snapshots started and found no renames or derivation changes.

You were right about the tests. One hand-built fixture on both sides proves nothing. Rebuilt it around the real serialized shape, with the noise that is actually there (weekendRate, ids, timestamps, all image variants) so the field names are pinned.

**Image keys**

You were right. The template resolves web_key, thumb_key, key, original_key (accommodationImage.js:55-69) and the digest had original_key first. It now imports resolveAccommodationImageKey, the same function the template uses, so the two cannot drift apart again. Added a test for a regenerated web variant.

**Pricing**

No extra guest fee exists, Property_Pricing only has cleaning and roomrate. Cleaning was in the digest but the site never renders it from the snapshot, it comes from the live quote (QuotePanel.jsx:90-93), so I dropped it. Minimum stay and check-in/check-out were missing though, both are rendered (buildWebsiteTemplateModel.js:292-298 and 315-323), so those are in now. Property type too, same reason.

**Nit**

Renamed to keepActivePropertyEntries.

**Also fixed while I was in here**

listingDetails was loaded once when the editor opened and never refreshed after publishing, so a listing change made in another tab could make the notice wrongly appear after an update and stay until reload. It is refetched after a publish now. And a failed staleness check on the builder page was cached as if it had passed, so a tab switch would not retry it. A failed check is no longer cached and shows no notice.

karimh01:
> The stale comparison seems to miss some listing fields that are actually rendered on the published site. buildWebsiteTemplateModel uses availabilityRestrictions (e.g. MinimumStay) and check-in/check-out values, but buildListingDigest() doesn’t include them. This means changing something like the minimum stay could leave the live site on the old value without showing the stale notice, and Update live site may remain disabled when there are no editor override changes. Could we include the remaining listing fields that affect the published template in the digest?

Good catch, you are right. availabilityRestrictions and checkIn were both rendered on the published site but missing from buildListingDigest, so a minimum stay change did not show the notice. Added both, and I dropped cleaning while I was in there since the site gets that from the live quote and not from the snapshot. Property type was missing for the same reason, that is in now too.

On the button: the gate is not override-only, it is canUpdateLiveSite = hasLiveSyncPending || liveSiteStaleness.isStale. But since the digest did not read the restrictions, isStale stayed false for a minimum stay change, so in practice the button did stay disabled exactly like you describe. With the digest fix it now enables.

The reason it reads as override-only is that the prop was still called hasLiveSyncPending while it carried canUpdateLiveSite. I split that into two props so each menu item reads the value its own handler checks. That also fixes Discard all changes, which looked enabled when only the listing was stale even though its handler returned early.

All pushed, could you take another look?

elcoleta (approved):

Issue #2560, opened 2026-02-16
Website: Host dashboard (Task Managment): Settings window


vanheerwaardenchiel-bit:
# ⚙️ Settings Window – User Stories & Checklist

Design mockups are uploaded in comments — implementation must strictly follow layout, spacing, hierarchy, colors, and interaction rules from the design.

---

# ✅ Quick Implementation Checklist (Dev Summary)

## General
- [x] Settings page accessible from **Tasks → Settings tab**
- [ ] Page displays **user account settings, team management, integrations, and task defaults**
- [x] Settings update **without full page refresh where possible**
- [x] Changes require **Save Changes confirmation**
- [x] Cancel button resets unsaved changes
- [ ] Later: Settings respect **role-based permissions**

---

# 📌 User Stories

---

#  Account Settings

## User Story
As a user, I want to manage my account information so that my contact details and credentials stay up to date.

### Acceptance Criteria

- [ ] Account section displays user profile information
- [ ] Profile image displayed
- [ ] Option available to **change profile photo**

Editable fields:

- [ ] Full Name
- [ ] Email
- [ ] Phone number

Password management:

- [ ] Password field displayed as hidden characters
- [ ] **Change Password** button available
- [ ] Clicking button opens password change flow

Additional behavior:

- [ ] Fields validate input format
- [ ] Email must be unique
- [ ] Changes apply only after clicking **Save Changes**

---

#  Notifications Settings

## User Story
As a user, I want to configure my notifications so that I only receive relevant task alerts.

### Acceptance Criteria

Notification panel includes toggle switches.

#### Email Notifications
- [x] Toggle **Task assigned to me**
- [x] Toggle **Task overdue**
- [x] Toggle **Task completed**

#### SMS Notifications
- [x] Toggle **Urgent tasks only**

#### In-App Notifications
- [x] Toggle **Enable notifications**

Additional behavior:

- [x] Toggles update settings instantly or on Save
- [x] Default settings applied for new users
- [x] Notification preferences stored per user

---

#  Default Property Task Settings

## User Story
As a property manager, I want to configure default task behavior so that tasks follow consistent operational rules.

### Acceptance Criteria

Settings include:

#### Default Task Priority
- [x] Dropdown selection
- [x] Options include:
  - [x] Low
  - [x] Medium
  - [x] High
  - [x] Urgent

#### Default Assignee
- [ ] Dropdown selection
- [ ] Options include:
  - [ ] Anyone
  - [ ] Specific team members

[...]

PR #3247, opened 2026-09-15
fix(account-settings): fix email errors, deferred saves, and form validation on Personal Data (#2560)

Nishchal-code:
## Close or Relate the Issue
- Closes #
- Related Issue: #2560

## Proposed Changes

Fixes to the host Personal Data settings form (validation and save behavior):

**Email (bug fix)**
- The backend correctly returns "This email address is already in use." for a duplicate email, but the frontend checked `response.ok` before checking the message, so that response was always masked by a generic "Failed to update email" alert. Fixed the check order.
- Replaced the email/verification flow's `alert()` popups with inline success/error messages under the field.

**Title & Sex (bug fix)**
- These were the only two profile fields saving immediately on change instead of waiting for "Save Changes," inconsistent with the rest of the form. Now deferred like every other field.

**Date of birth (new feature)**
- Replaced the manual digit-typing masked input with an actual calendar picker (`react-datepicker`, already used elsewhere in the app for booking dates), with year/month dropdowns for quick navigation and `maxDate` so the picker itself won't allow a future date.

**Sex options (new feature)**
- Added "Other" and "Prefer not to say" to the dropdown, alongside Female/Male.

**Name validation (bug fix)**
- First/last name previously accepted anything (digits, symbols, unlimited length). Now validated for format (letters/spaces/hyphens/apostrophes) and length, matching the pattern already used for nationality.
- Fixed `saveUserPhone`'s error handler firing the same failure alert twice (copy-paste bug).

**Not included here:** the profile photo upload fix is a separate PR (linked below) since it's an independent Lambda change unrelated to this form's validation logic.

**Known gap (not in this PR):** none of Email/Name/Phone/DOB/Title/Sex/Place of birth/Nationality validate server-side beyond what the underlying Cognito/Lambda endpoints already enforce. Flagged as follow-up work.

## Change Type
- [x] Bug fix
- [x] New feature

## Change Size

## Refactoring
N/A — no pure refactors; behavior fixes/additions each paired with test changes in the same commit.

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [x] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [ ] Conventions
- [ ] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


karimh01 (changes requested):
The new profile-photo Lambda is good: username comes from the token, not the body, so you can only touch your own photo. That's the right way to do what the old frontend code did with an AWS key.

But this PR is a lot more than the title says. A new Lambda with shared exception classes, plus nine frontend commits that each do something different: photo upload, success feedback, email errors, deferred saves, a date picker, extra sex options, name validation, a removed alert. 928 lines across 28 files.

If the date picker breaks next week, someone has to revert a PR that also contains the photo Lambda. That's the problem with bundling.

I'd split it: the Lambda and the upload fix together, since they belong to each other. The rest in one or two PRs for the form changes. And a title that says what's in it.

Not blocking on the code itself, but the packaging needs work.

Nishchal-code:
> The new profile-photo Lambda is good: username comes from the token, not the body, so you can only touch your own photo. That's the right way to do what the old frontend code did with an AWS key.
> 
> But this PR is a lot more than the title says. A new Lambda with shared exception classes, plus nine frontend commits that each do something different: photo upload, success feedback, email errors, deferred saves, a date picker, extra sex options, name validation, a removed alert. 928 lines across 28 files.
> 
> If the date picker breaks next week, someone has to revert a PR that also contains the photo Lambda. That's the problem with bundling.
> 
> I'd split it: the Lambda and the upload fix together, since they belong to each other. The rest in one or two PRs for the form changes. And a title that says what's in it.
> 
> Not blocking on the code itself, but the packaging needs work.

Thank you for useful feedback!!!!

Rakib730066 (dismissed):

karimh01 (changes requested):
Packaging is sorted, thanks for splitting the photo Lambda into its own PR. This reads as one thing now.

And good catch on the save-all flow. Promise.allSettled resolving even when a save swallowed its own error is exactly the kind of thing that looks fine in tests and lies to the host.

There are four points Copilot collapsed that nobody picked up. Two of them I would fix before this merges.

**The inline messages are silent for screen readers.** You replaced alert() with plain paragraphs. An alert announces itself, a <p> does not, so a host using a screen reader submits an email and hears nothing at all. That is a step back for those users even though the visual feedback is better. The error needs role="alert" and the success role="status", ideally tied to the field with aria-describedby.

**The date picker can be clipped.** personal-data-card has overflow: hidden (hostSettings.css:219-224) and the default DatePicker popper mounts inside the card, so opening the calendar near the bottom of the form cuts it off. The booking date pickers already solve this with the datepicker-portal and fixed positioning, so the pattern is there.

Two smaller ones, your call:

pd-field is a flex row, so the feedback paragraphs render beside the input instead of under it, and the 136px left margin does not make a flex item wrap. A long duplicate-email error will squeeze the input.

And showEmailSuccess leaves emailSuccess true for 2.5 seconds, while the change handler only clears emailError. So editing the address in that window still shows "Email updated!" under a different unsaved value.

One thing worth knowing rather than fixing: Sonar reports 0.0% coverage on new code even though there are three test files in the diff. Might be worth checking whether those paths are actually being measured.

karimh01 (changes requested):
All my earlier points are fixed. The error messages have role="alert" and the success message role="status", and almost all of them are linked to their field with aria-describedby. The date picker now renders in the portal with fixed positioning, the same way the booking date pickers do it. The messages sit under the field instead of next to it, and emailSuccess is cleared when the address changes, with a test.

One new problem, and I think it came from merging acceptance. The photo feedback from #3249 is gone again. On acceptance PersonalDataForm.jsx shows photoSuccess for "uploaded" and "removed", and this branch removes those two lines, so uploading or removing a photo gives the host no confirmation at all. The .pd-photo-error and .pd-photo-success styles are removed from the CSS too, so photoError now falls back to the field styling with margin-left: 136px and gets squeezed inside the 200px photo column. Bringing back the three lines and the two CSS blocks from acceptance fixes it.

Two smaller things, your call:
- When the email only still needs verification, the page already shows "Saved!", because the save chain returns true as soon as verification mode starts.
- Email errors are now inline, but name, phone and date of birth still use alert(), so the page mixes two ways of showing errors.

PR #3249, opened 2026-09-15
fix(account-settings): fix broken profile photo upload (#2560)

Nishchal-code:
## Close or Relate the Issue
- Closes #
- Related Issue: #2560

## Proposed Changes

**Profile photo (bug fix)**
- Root cause: upload was calling an unrelated chat-attachment Lambda pointed at a private bucket, so photos either failed to upload or saved a URL that never displayed.
- Built a dedicated `profile-photo` Lambda that validates the image and writes it to the existing public `accommodation/images/` S3 prefix; frontend now sends the photo as base64 with the user's access token — the username is taken from the verified Cognito token, not the request body, so a user can only ever touch their own photo.
- Upload/Remove buttons now show proper "Uploading…"/"Removing…" status (previously stuck on "Working…", and Remove showed nothing).
- Added success feedback ("Photo uploaded!" / "Photo removed!") after each action.
- Two follow-up commits address a SonarQube duplication gate failure: moved two exception classes to `.shared` (matching this repo's existing convention), and restructured `index.js`'s routing boilerplate to remove repeated CORS-header literals.

**Split from #3247:** this PR was split out from the original combined PR so the photo Lambda can be reverted independently of the unrelated Personal Data form fixes, which are in #3247.

## Change Type
- [x] Bug fix

## Change Size
- [x] Medium change (less than 500)

## Refactoring
Refactored `backend/functions/profile-photo/index.js` and moved two exception classes to `backend/functions/.shared/` to resolve a SonarQube duplication gate failure — no behavior change, verified against the original response shapes.

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [x] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [ x] PR title is descriptive and includes issue number
- [ ] Conventions
- [ ] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


karimh01 (changes requested):
Looks good, two things to fix first.

**A big photo gives the wrong message**

The browser stops files over 5 MB and the Lambda stops decoded images over 5 MB. But base64 makes a file about a third bigger, and AWS caps a Lambda request at 6 MB. So anything over about 4.4 MB gets past the browser check, API Gateway blocks it before your code runs, and the host just sees "Failed to upload photo. Please try again."

Same for your own errors: profileUpload.js:17-19 throws a fixed string on any non-2xx and never reads the body, so "Unsupported image type." and "Image must be 5MB or smaller." never show up. I would set the browser limit to about 4 MB, say that in the text, and show the server message on a 4xx.

**Old photos stay forever**

Every upload gets a new UUID, so the previous photo keeps sitting in the public bucket. Not a blocker.

**What I checked and is fine**

The username comes from Cognito, the body is only used for the image, and the key is username plus UUID. So nobody can write or overwrite someone else's photo.

Validation only looks at the label, not the actual bytes, so someone logged in could store non-image content as a .png. SVG is excluded and browsers will not run it, so it is junk in a bucket more than a real risk. PropertyHandler already uses sharp for this if you want it later.

**The .shared import depth is right**

It points at nothing in the repo, but deploy.yml puts .shared one level above the function files in the zip, so it works in AWS. Same as every other Lambda.

Good to know: jest.config.mjs maps every depth from one to four to the same folder, so a wrong one passes the tests and only breaks in AWS. I hit that myself a few weeks ago.

Rakib730066 (approved):

karimh01 (approved):
Thanks for splitting this out, it reviews a lot easier on its own.

Both points from last time are fixed. The limit is 4 MiB on the client and the same 4 MiB on the decoded bytes in the Lambda, and a 4 MiB file ends up around 700 KB under the 6 MiB ceiling, so nothing that passes the client check can be cut off by API Gateway anymore. And the host now sees the real server message instead of the generic one. Tests cover both.

Username still comes from the token only, and the .shared depths are right for the deployed Lambda.

Two small things before it merges: the branch has a conflict in frontend/web/src/components/settings/constants.js, and profileUpload.js:23 has a comment, and there should be no comments in the code.

One thing I noticed that I did not raise last time, so not blocking this PR: Remove only clears the Cognito picture attribute and the object stays in S3, where the bucket policy gives public read on images/*. So a host who removes his photo because he wants it offline still has it reachable at the old URL. Together with the old photos piling up, worth a follow-up that deletes under the user's own images/profile/<username>/ prefix. That prefix comes from the token, so it can never touch someone else's photo. The Lambda role would need s3:DeleteObject on that path.

PR #3250, opened 2026-09-15
#3106 Invalidate each CloudFront tenant on deploy

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3106

## Proposed Changes
Description:
The "Deploy direct booking sites" workflow was failing on every run. The last step called create-invalidation on the direct booking sites distribution, but that distribution has ConnectionMode tenant-only. A tenant-only distribution does not serve anything itself, only its tenants do, so CloudFront answers NoSuchDistribution and the deploy fails.

The step now lists the tenants of the distribution and creates an invalidation for each one, with the same paths as before. If there are no tenants yet it writes a notice and passes instead of failing.

**Console work already done**: the deploy user Github-Lambda-Deployer had no access to the sites bucket at all, so the S3 sync step was failing too. I added an inline policy DirectBookingSitesDeploy with s3:PutObject, GetObject, DeleteObject and ListBucket on domits-direct-booking-sites-acceptance, plus cloudfront:CreateInvalidation, ListDistributionTenants and CreateInvalidationForDistributionTenant. The CloudFront actions are on resource "*" because AWS does not support resource level permissions for them.

**Nothing changes in production**, this only touches the acceptance deploy workflow.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
Nothing refactored.

## Evidence
The S3 sync step is green after the policy change, so the bundle does reach the bucket now. The invalidation step cannot be proven green until this is merged, because the workflow only runs on acceptance. I checked the distribution is tenant-only with get-distribution, and confirmed both CLI commands exist and take the arguments used here.

## Responsiveness
No UI.

## Npm Packages

## Checklist
- [x] Conventions
  - [x] No config files
  - [x] No global styling
  - [x] No console.log
  - [x] No commented-out code
- [x] Testing
  - [x] Code tested locally
  - [ ] Jest tests are included
  - [ ] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


elcoleta (approved):
Looks good. The workflow now handles the tenant-only CloudFront setup correctly by discovering the tenants for the distribution and invalidating each tenant individually, while also handling the no-tenant case cleanly. The caller reference is unique per workflow run/attempt and tenant as well. Approving.

BeMounir (approved):

EnesVeli (approved):
Approving

PR #3251, opened 2026-09-15
#3106 Store the domain claim before creating the CloudFront tenant

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3106

## Proposed Changes
Description:
I tested the custom domain flow end to end on acceptance and found that connecting a domain can never work for a normal host. CloudFront refuses CreateDistributionTenant unless the domain already resolves to a CloudFront resource. The exact error is "Could not verify Domain Name ownership". I ran the same command twice, once before adding the CNAME and once after, and only the second one worked.

Our flow did it the other way around. The host typed a domain, we created the tenant, and only then showed them the CNAME. Because the tenant creation threw, nothing was stored and the host never saw the CNAME, so retrying could never help.

Connect now stores the domain first with status PENDING, reason dns_required and the CNAME instruction filled in, without calling CloudFront. The tenant is created on the next sync, which runs when the panel loads and when the host presses Check again. The ownership error keeps the record on dns_required instead of failing, so the host can fix their DNS and try again.

Two things that come with this:
- A domain another site holds is always domain_taken, whether that site has a tenant or not. I first let a new claim take over a record without a tenant, to stop someone reserving a domain they do not own. Enes pointed out that this traded squatting for hijacking: the CNAME target is the same for every site and CloudFront only checks that the domain resolves to CloudFront, so a second site could claim the domain after the real owner set up their DNS, create a tenant on it, and serve its own site there with a valid certificate. That is worse than squatting, so the takeover is out. Squatting on an unproven claim is a known gap now, with an expiry for tenant-less rows or a per-site TXT record as the follow-up.
- The panel tells the host to add the CNAME and then press Check again.

**Known gap**: once a domain is stored the input field disappears, so a host who typed the wrong domain cannot change it. Removing a domain is a separate PR, that will fix it.

**On the prefix-aware ownership check:** a tenant name that starts with dbw- must be exactly dbw-<siteId>, but a name without that prefix is treated as legacy and allowed. There is one such tenant, developers-test, created by hand before the naming convention existed. CloudFront tenant names cannot be renamed, so a strict rule would mean recreating that tenant and churning the certificate on developers.domits.com. The race this closes always hands a site a dbw- tenant belonging to another site, so the exception does not weaken the fix. Tightening it to strict is a one-line change once developers-test is recreated.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
Tenant creation moved out of requestCustomDomain into a new provisionTenant method that the sync calls. Same code, different place.

## Evidence
Tested on acceptance with my own domain. test.cablets.nl now points at CloudFront, the certificate was issued and applied, and the site is served over https. That proved the chain works and showed the ordering problem: the tenant call failed before the CNAME existed and succeeded after.

## Responsiveness
No UI changes beyond copy.

## Npm Packages

## Checklist
- [x] Conventions
  - [x] No config files
  - [x] No global styling
  - [x] No console.log
  - [x] No commented-out code
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included (48 in the two PropertyHandler domain test files)
  - [x] Jest tests are


elcoleta (approved):
The takeover logic looks correct for the normal sequential case, but I think there is a TOCTOU race here. We first read existingRecord and allow takeover when it has no tenantId, then ensureDomain() later performs an unconditional ON CONFLICT (domain) DO UPDATE.
If the original site provisions its tenant between those two operations, the new claimant can still overwrite the row even though the domain has just become proven. The reverse interleaving is also possible: the row can be reassigned first, then the original sync updates that same row by id with its tenant details.
Could the claim transfer be made atomic/conditional in the repository so it only reassigns the domain if it still has no tenant at update time?

karimh01:
> The takeover logic looks correct for the normal sequential case, but I think there is a TOCTOU race here. We first read existingRecord and allow takeover when it has no tenantId, then ensureDomain() later performs an unconditional ON CONFLICT (domain) DO UPDATE. If the original site provisions its tenant between those two operations, the new claimant can still overwrite the row even though the domain has just become proven. The reverse interleaving is also possible: the row can be reassigned first, then the original sync updates that same row by id with its tenant details. Could the claim transfer be made atomic/conditional in the repository so it only reassigns the domain if it still has no tenant at update time?

Good catch, and it turned out worse than a stale write. Both interleavings you named are real, and the second one is the bad one.

If the row gets reassigned to site B first and site A's sync then calls updateDomainStatusById by id only, A's tenant id lands on B's row. B now holds a proven claim on A's tenant. A's panel shows the connect form again, and re-entering its own domain answers domain_taken, so A is locked out of both its domain and its tenant. With the remove flow from my other PR, B could then disable and delete A's tenant, since every tenant call keys purely on the stored tenantId and never checks the tenant belongs to that site.

It also does not need luck. Any host can re-post another host's unproven domain repeatedly, and each post resets the row, so A can be kept from ever provisioning even without hitting the race.

Fixed both sides:

The takeover is now a compare-and-swap. New repository method claimDomain does UPDATE ... WHERE domain = $1 AND site_id = $2 AND updated_at = $3 with the site id and updatedAt from the read. No row matched means someone got there first, and the service answers domain_taken. Nothing inside the JSON column is compared, so it stays portable on DSQL. A brand new domain still goes through the plain insert.

Every status write is scoped to the site. updateDomainStatusById and updateDomainVerificationDetailsById now take a siteId and update WHERE id = $1 AND site_id = $2. In provisionTenant a null result means the row changed owner, so it disables the tenant it just created and answers domain_taken rather than leaving an enabled tenant that could serve under someone else's row.

Tests for both: a takeover that loses the race answers domain_taken and touches nothing, and a provisioning write on a row that changed owner throws and disables its tenant.

Two follow-ups I have noted: the same site_id predicate on deleteDomainById and the removal writes in the remove PR, and a tenant ownership check comparing tenant.name with dbw-<siteId> before any disable, delete or certificate call. That last one closes the whole class even if a row ever gets mis-assigned some other way.

karimh01:
Also added the tenant ownership check I mentioned as a follow-up. Tenant names are dbw-<siteId>, so the name is the proof of ownership. Before the service applies a certificate, verifies DNS or adopts a tenant, the name has to be ours. A dbw- name belonging to another site is refused with tenant_not_owned (409), the row records lastError tenant_not_owned, and nothing is written at CloudFront.

Kept it in this PR because it closes the same hole from the other side: even if a row somehow ends up with a tenant id that is not ours, the tenant itself still cannot be touched.

One exception, explained in the description: a tenant name without the dbw- prefix is legacy and allowed, because of developers-test.

EnesVeli (approved):
This is careful work, and the ordering bug is well explained. The follow-up commits close Achraf's race: claimDomain compares updated_at and all writes are scoped to the site, both orders are tested (151, 249, 263), and a single conditional UPDATE is the right approach for DSQL.

One blocking issue: a domain can be taken over while the real owner is setting up DNS.

The CNAME value is config.routingEndpoint (websiteCustomDomainService.js:168), which is the same for every site, and CloudFront's check only proves that the domain points at CloudFront. So:

Site A claims www.example.com and gets dns_required with no tenant (245-260).
A adds the CNAME.
Before A syncs, site B requests the same domain. There's no tenant, so no DOMAIN_TAKEN (228), and takeOverUnprovenDomain moves the row to B (257-259).
B's sync creates dbw-<B>. CloudFront accepts it because the domain already points at CloudFront, thanks to A.
B serves its site on A's domain with a valid certificate, and A is locked out because the row now has a tenant (228).

The DNS proof isn't tied to a site, so whoever syncs first after DNS resolves wins, and a site can also keep taking over a pending claim to stop the owner from ever provisioning. The shared CNAME isn't new, but this PR makes the unproven-but-DNS-ready state a normal step and deliberately makes it takeoverable.

Suggestion: require a per-claim token (for example TXT _domits-verify.<domain>) before provisionTenant, and before a takeover. If the team accepts the risk for now because custom domains aren't public, let's get a senior to confirm and track it in an issue.

Non-blocking

When the race is lost, the fresh tenant is disabled, not deleted (305). If a disabled tenant still holds the domain, the winner hits CNAMEAlreadyExists → FAILED / domain_in_use_elsewhere (197-198) and is never re-synced (80-81). Do you know how CloudFront treats disabled tenants here?
ensureDomain still does an unconditional upsert for new claims (260). The risk is low since the row can't have a tenant yet, but ON CONFLICT DO NOTHING plus a re-read would match claimDomain.
Concurrent writes to the same row fail with DSQL 40001 and come back as a 500, not DOMAIN_TAKEN.
If disableTenant throws (305), its error replaces DOMAIN_TAKEN and the tenant stays enabled.

karimh01:
> This is careful work, and the ordering bug is well explained. The follow-up commits close Achraf's race: claimDomain compares updated_at and all writes are scoped to the site, both orders are tested (151, 249, 263), and a single conditional UPDATE is the right approach for DSQL.
> 
> One blocking issue: a domain can be taken over while the real owner is setting up DNS.
> 
> The CNAME value is config.routingEndpoint (websiteCustomDomainService.js:168), which is the same for every site, and CloudFront's check only proves that the domain points at CloudFront. So:
> 
> Site A claims [www.example.com](http://www.example.com) and gets dns_required with no tenant (245-260). A adds the CNAME. Before A syncs, site B requests the same domain. There's no tenant, so no DOMAIN_TAKEN (228), and takeOverUnprovenDomain moves the row to B (257-259). B's sync creates dbw-. CloudFront accepts it because the domain already points at CloudFront, thanks to A. B serves its site on A's domain with a valid certificate, and A is locked out because the row now has a tenant (228).
> 
> The DNS proof isn't tied to a site, so whoever syncs first after DNS resolves wins, and a site can also keep taking over a pending claim to stop the owner from ever provisioning. The shared CNAME isn't new, but this PR makes the unproven-but-DNS-ready state a normal step and deliberately makes it takeoverable.
> 
> Suggestion: require a per-claim token (for example TXT _domits-verify.) before provisionTenant, and before a takeover. If the team accepts the risk for now because custom domains aren't public, let's get a senior to confirm and track it in an issue.
> 
> Non-blocking
> 
> When the race is lost, the fresh tenant is disabled, not deleted (305). If a disabled tenant still holds the domain, the winner hits CNAMEAlreadyExists → FAILED / domain_in_use_elsewhere (197-198) and is never re-synced (80-81). Do you know how CloudFront treats disabled tenants here? ensureDomain still does an unconditional upsert for new claims (260). The risk is low since the row can't have a tenant yet, but ON CONFLICT DO NOTHING plus a re-read would match claimDomain. Concurrent writes to the same row fail with DSQL 40001 and come back as a 500, not DOMAIN_TAKEN. If disableTenant throws (305), its error replaces DOMAIN_TAKEN and the tenant stays enabled.

You are right, and it is not even a race. B does not have to be fast, only has to know the domain and claim it before A opens the panel again. That can be hours.

I traced it exactly as you describe: A's row has no tenant so the domain_taken guard lets B through, the takeover moves the row, and B's createTenant succeeds because A's CNAME already points the domain at CloudFront. CloudFront's check is "points at CloudFront", never "points at you". The managed certificate validates over HTTP at the routing endpoint, which CloudFront answers itself, so B ends up serving its site on A's domain with a valid certificate. My tenant ownership check does not help here, since B's tenant is genuinely named for B.

The takeover was meant to stop squatting. It traded squatting, where the victim keeps control and the squatter gets nothing, for hijacking, where the attacker gets the domain. That is clearly worse, so I took the takeover out. A domain another site holds now always answers domain_taken, tenant or not. That also made claimDomain dead, so the repository method is gone too. The compare-and-swap protected exactly the transition we no longer allow.

What stays: the site-scoped status writes, the disable-and-refuse when a row changes owner during provisioning, and the tenant ownership check.

Squatting is now a known gap again. I wrote it up in the design pack with the two options: an expiry for tenant-less rows, or a per-site DNS proof, a _domits.<domain> TXT record with the site id, checked before the tenant is created. The second closes it completely and matches how AWS handles the shared endpoint case with _cf-challenge. That is its own PR.

On your non-blocking point about the disabled loser tenant: you are right and it is worse than cosmetic. AWS says a domain with a pending certificate request cannot be associated with any other CloudFront resource, and disabling a tenant is not listed as something that cancels that request. A tenant created seconds earlier always has one pending. So the winner hits CNAMEAlreadyExists, lands on FAILED with domain_in_use_elsewhere, and stays blocked until someone deletes the orphan by hand. Same cleanup as the orphan issue I opened today: on CNAMEAlreadyExists, look the tenant up by domain, and if it is ours, disabled and deployed, delete it and retry.

The other two I will pick up: ensureDomain doing an unconditional upsert for new claims, and a DSQL 40001 on concurrent writes coming back as a 500 instead of domain_taken.

PR #3252, opened 2026-09-16
feat(inbox): add manual mark read and unread (#2291)

elcoleta:
## Close or Relate the Issue
- Closes #

- Related Issue: #2291

## Proposed Changes
Description:

Adds manual **Mark as read / Mark as unread** controls to the Unified Inbox.

Backend:
- Adds support for `POST /threads/{id}/unread`
- Authorizes access before updating the conversation
- Marks only the latest incoming message for the authenticated user as unread
- Uses deterministic ordering when selecting the latest message
- Keeps the existing Mark as read behavior unchanged

Frontend:
- Adds a visible three-dot action menu to each conversation
- Shows **Mark as read** for unread conversations
- Shows **Mark as unread** for read conversations
- Keeps the menu attached to the correct conversation row
- Fixes row targeting for Close chat
- Handles the case where a conversation is opened while Mark as unread is still in progress

The new `/unread` backend route is implemented and tested, but still needs to be available in the deployed API Gateway environment for full end-to-end testing.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring

N/A

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [x] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

No NPM package changes were made.

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

Testing completed:
- Unified Messaging backend: 392/392 tests passed
- Frontend messaging: 67/67 tests passed
- ContactList: 22/22 tests passed
- ESLint: 0 errors
- `git diff --check`: clean
- Mark as read manually verified end-to-end

## Keep or delete my branch
- [ ] Keep my branch


karimh01 (changes requested):
Checked the code against the Copilot points, two of them do not hold. Authorization is solid: both paths go through getAuthenticatedUser and assertThreadAccess, and the unread query is scoped to recipientId, so a participant can only flip a message addressed to themselves.

**The 401 is invisible to the host**

Worth knowing before the route is added: right now the request does reach the Lambda, through the catch-all /{proxy+} with ANY. But that one has authorization NONE, while /threads/{id}/read has the Cognito authorizer. So the call arrives without claims and the Lambda answers 401.

On screen nothing happens. The menu closes before the request goes out, markThreadUnread throws on the 401, and the catch in handleMarkAsUnread is empty (ContactList.jsx:357-359). No message, no badge change. The row just stays as it was.

Two things from that. The new route needs the same Cognito authorizer as /read, otherwise it keeps failing after merge. And an error notice in that catch, and in the matching read handler, would stop this from being silent.

**The compensating read can leave the row wrong**

Copilot is right on this one. In the branch where the host opened the thread while the request was in flight, markThreadUnread succeeds, markThreadRead throws, the catch swallows it and no local update runs (ContactList.jsx:341-359). The row keeps unreadCount 0 while the backend now has one unread message. It fixes itself on the next contacts fetch, so it is small, but it is real.

**Menu on threadless contacts**

Also right, but harder to hit than it sounds. Every host contact from useFetchContacts has a threadId. The only way to get one without is a locally created row from an incoming socket message (ContactList.jsx:132). So the dead menu item exists in that edge case only.

**No repository test for the new query**

The only repository test in this Lambda is the automation one. Nothing covers the threadId and recipientId predicates or the createdAt DESC, id DESC ordering, and the service tests mock both calls. Worth adding since the ordering is the whole point of picking the latest incoming message.

**Two Copilot points that do not hold**

The affected-row count is not wrong. Mark as unread only shows when unreadCount is 0, so everything incoming was already read, and flipping one makes the true count exactly 1. The two numbers agree by construction. Fragile way to express a count, but not a bug today.

The menu is not clipped either. It is absolutely positioned in a row that is overflow visible, inside a list that is overflow auto, and nothing above it clips. A menu on the last row does drop below the fold and needs a scroll, which is a bit awkward, but nothing gets cut off.

EnesVeli (dismissed):
Backend looks good: assertThreadAccess plus the recipientId-scoped query means only your own incoming message can be flipped. The fix in 925c2dca0 for a failed compensating read works.

Still blocking (from Karim's review):

Both handlers still swallow errors (ContactList.jsx:328-330 and 360-362), and the menu closes before the request. Until the /threads/{id}/unread route exists in API Gateway with the Cognito authorizer, every click will fail with no feedback. Please show an error message on failure, and make sure the route plus authorizer are tracked before this merges.

Correction on threadless contacts: this isn't only socket-created rows. For hosts, useFetchContacts also merges the legacy contacts API (useFetchContacts.js:360-386), and normalizeLegacy (299-314) doesn't set a threadId. So any host with legacy contacts gets menu items that silently do nothing. Please hide the mark actions when there's no threadId.

Non-blocking:

On mobile, Messages.js:409 unmounts ContactList when a thread opens, so activeThreadIdRef never updates and the in-flight correction at line 347 can't run. The opened thread can end up unread.
result.updated is a row count, not the unread total. If a message arrives over the websocket while the request is in flight, the badge is set to 1 when the real count is 2.
The menu is keyed by threadId || partnerId (310, 420-422), so two legacy rows with the same guest open the menu together.
On a thread with no incoming messages, "Mark as unread" returns updated: 0 and nothing visible happens. A short message would help.
A repository test for the threadId/recipientId filters and the createdAt DESC, id DESC ordering is still missing, as Karim noted.

Mehedihasanb (changes requested):
The API Gateway gap for `/threads/{id}/unread` is a deploy step, not a code issue, and I wouldn't block on it by itself — same pattern as other PRs shipping code ahead of console/infra config. Worth tracking so it doesn't get forgotten between merge and the route actually existing.

Three things I'd want addressed first:

1. The action menu shows for every contact row, including legacy ones with no `threadId`. Both `handleMarkAsRead` and `handleMarkAsUnread` just return early when `threadId` is missing — no error, no feedback, the click does nothing. Can we hide the mark actions (or the whole menu) for rows without a thread?

2. The description says this handles a conversation being opened while "Mark as unread" is still in flight, via `activeThreadIdRef`. Copilot's latest pass says that doesn't work on mobile, since `ContactList` unmounts when a contact opens there, so the ref never sees the new active thread and the corrective re-read never fires — the conversation the host is actively viewing stays marked unread. Can you confirm whether this reaches mobile, and either fix it or scope the PR description to desktop only?

3. `result.updated` (rows affected by flipping the one latest message) gets written straight into the row's `unreadCount` badge. Looks right in the common case, but it's really "did this one write happen," not "how many messages are unread." Under real concurrency (another message landing while the menu was open) the badge could undercount. Not asking for a fix today, just flagging it as a simplification worth being aware of rather than an accurate count.

The two CSS notes (menu possibly clipped by the scrollable list's overflow, and `all: unset` dropping focus visibility with no `:focus-visible` replacement) are minor and Copilot's suggested diff on the second one looks like a straight take.

Backend side looks solid — deterministic ordering on `getLatestIncomingMessage`, the authorization check reuses the existing tested pattern, and the route regexes don't collide (`/read$` vs `/unread$`, checked the "un" doesn't accidentally match "/read" at the end).

karimh01 (changes requested):
The authorization in the code is solid. Both operations call assertThreadAccess (messageService.js:180), read updates require the authenticated recipient (messageRepository.js:203), and unread picks the latest message by thread and recipient (:228), with the message id coming from that stored result and not from the request. So nobody can change the read state of someone else's thread. The unread count also handles a message arriving mid-flight: on success it adds to the current count instead of overwriting it, and that is tested.

Both points from last time are still open, plus one new one.

**Mark as unread still answers 401.** I checked the live API and the deployed stage. POST /threads/{id}/read has its own resource with the Cognito authorizer, but /threads/{id}/unread has no resource, so it falls through to /{proxy+} with authorization NONE. The new Lambda route at UnifiedMessaging/index.js:139 cannot fix that, because the controller needs authorizer claims and the proxy does not provide them. It needs its own resource with the same authorizer as read, OPTIONS with Access-Control-Allow-Methods 'OPTIONS,POST', and a stage deploy.

**Failures are still invisible.** ContactList.jsx:351 and :382 catch the error without showing anything, and the menu has already closed, so the host cannot tell a failure from a success. The same goes for the corrective read failure at :373. Show an error and let them retry, and keep the current count when it fails.

**Mark as read can wipe the count of a newer message.** ContactList.jsx:349 clears the count after any successful response through the helper at :173. If a new message arrives after the read update committed but before the response comes back, the response resets the count to zero. That message stays unread in the database, and in a conversation that is not open nothing corrects it. The unread path already solves this by adding to the current count; read needs the same kind of care, for example by only clearing what the read actually covered.

**Tests.** The unread failure path is not tested: the corrective-read failure test first lets markThreadUnread succeed. That is the case that happens in production today. The read race above has no test either.

Small: the PR adds comments, for example ContactList.jsx:32 and :178, the catches, and ContactList.test.js:461. There should be no comments in the code.

karimh01 (changes requested):
Four of the five points are fixed properly.

Failures now show a notice on all four paths, including the corrective read with its own wording, and the toast container is mounted at the app root so it actually renders on the inbox. Mark as read no longer wipes a newer message: it subtracts the confirmed updated count instead of setting zero. The tests are real. The unread failure is tested, and the race test holds markThreadRead open, injects a message through the real WebSocket provider, and checks the count stays at 1. Against the old code that test would fail. The comments are gone too. And nobody can change someone else's read state, both operations check thread access and scope the query to the caller.

**Mark as unread still answers 401.** I checked the live API again. Under /threads there is /threads, /threads/{id} and /threads/{id}/read, but no unread resource. So the request still falls through to /{proxy+} with authorization NONE, and authContext.js:39-41 throws because there are no claims. The default stage was last deployed on 16 September, so nothing was deployed for this either. It needs a /threads/{id}/unread resource with POST using the same Cognito authorizer as read (wgzgpr), OPTIONS with Access-Control-Allow-Methods matching the read route, and a stage deploy. None of that is in the repo, which is why the tests stay green while the button does not work.

Two small things, not blocking:

When the server answers updated: 0, local state is now left alone. That is the more correct reading, but a badge that ever drifts above the server's count will now stay wrong instead of resetting on the next read.

markMessageUnread(messageId) takes only an id with no owner scope. It is safe now because the id comes from getLatestIncomingMessage for the caller, but the method on its own would flip any message it gets. A scope on the update would keep it safe if it ever gets a second caller.

karimh01 (approved):
Checked the API Gateway side with the CLI: /threads/{id}/unread now exists on UnifiedMessaging with POST behind the same Cognito authorizer as /read, plus OPTIONS, and both are in the live default stage. The code was already good in my last review, so this was the last thing missing. The full end-to-end test can only happen after the merge when the Lambda goes live, so do that test right after merging.

EnesVeli (approved):

Mehedihasanb (approved):

PR #3253, opened 2026-09-16
#3106 Let the host remove a custom domain

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3106

## Proposed Changes
Description:
A host could connect a custom domain but never remove it. Once a domain was stored the input field disappeared, so a host who typed the wrong domain was stuck with it.

Removing is two steps at CloudFront, not one. DeleteDistributionTenant answers ResourceNotDisabled while the tenant is still enabled, so you have to disable it first, wait for that to roll out, and only then delete. I measured it on a real tenant and the rollout took about fifteen seconds.

How it works for the host: pressing Remove domain moves the row to a new REMOVING status and disables the tenant. The panel keeps showing the domain with a line saying it is being removed and to press Check again in a moment. On the next sync, which runs on panel load and on Check again, the tenant is deleted and the row is removed, and the panel goes back to the connect form.

The status is written before the CloudFront call on purpose. That way a failed write can never leave a disabled tenant behind an ACTIVE row, and if the disable itself fails the next sync retries it.

The remove request carries the domain the host is looking at, not just the site id. If that no longer matches what is stored, the request is refused with domain_not_found and the panel reloads. Without that, a stale tab could remove a domain the host connected in the meantime.

Edge cases handled: a row that never got a tenant is just deleted, and a tenant that is already gone at CloudFront counts as removed.

**Console work after merge**: API Gateway needs a DELETE method on /property/website/domains, and the Lambda role needs cloudfront:DeleteDistributionTenant.

**Known limit**: if a host presses Remove and never opens the panel again, the tenant stays disabled and the domain stays claimed in our table until someone opens it. There is no background job, same as the rest of the domain sync.

**Separate issue**: deleting a website that has a custom domain attached removes the row but leaves the tenant at CloudFront. The domain then shows an unavailable page and cannot be connected to a new site. That is not new in this PR, I will open an issue for it.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
Nothing refactored.

## Evidence
Tested the two steps by hand on a real tenant with the AWS CLI: disabling took about fifteen seconds to reach Deployed, and deleting only worked after that. Trying to delete an enabled tenant gives ResourceNotDisabled, which is what this flow is built around.

## Responsiveness
One extra button in the existing button row and a line of copy.

- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [x] UI checked on desktop

## Npm Packages

## Checklist
- [x] Conventions
  - [x] No config files have been added
  - [x] No global styling
  - [x] No console.log left in code
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included, 20 across the backend domain tests and the frontend panel, hook and service tests
  - [x] Jest tests are passing, 104 backend and 57 frontend

## Keep or delete my branch
- [ ] Keep my branch


elcoleta (approved):
Looks good. The removal flow correctly handles CloudFront’s disable-before-delete requirement, persists REMOVING before making the CloudFront change, and safely continues the cleanup on subsequent syncs. The stale-domain guard and the no-tenant/already-deleted cases are also covered well by the tests. Approving.

karimh01:
Heads up for whoever reviews this: #3251 changes the signature of updateDomainStatusById and updateDomainVerificationDetailsById, both take a siteId now. This branch calls both.

Two of those will show up as merge conflicts, so they get caught. But markDomainRemoving merges silently and then breaks on every Remove click, because with the new signature the status string lands in the siteId slot and normalizeDomainStatus throws.

I will fix that here once #3251 is merged.

EnesVeli (approved):
The removal flow is solid: disable before delete matches CloudFront's ResourceNotDisabled behaviour, REMOVING is written before the CloudFront call, and the stale-tab domain check is a good guard.

Blocking: this PR and #3251 break removal once both are merged.

[#3251](https://github.com/domits1/Domits/issues/3251) changes updateDomainStatusById to (domainId, siteId, status, verificationDetails). This PR calls it as (record.id, DOMAIN_STATUS.REMOVING, details) in markDomainRemoving. After both merge, siteId is "REMOVING" and status is the details object, so normalizeDomainStatus throws and every removal of a domain with a tenant returns domain_remove_failed.

I ran git merge-tree on both branches: the service conflicts only in syncCustomDomain, and markDomainRemoving merges silently. The test mock uses the old argument order (websiteCustomDomainService.test.js:40, assertion at 526), so CI stays green. Either merge order breaks it. Same situation as #3228 and #3181 last week.

Could you rebase this on [#3251](https://github.com/domits1/Domits/issues/3251), pass site.id in markDomainRemoving, and update the mock and assertion?

Non-blocking

deleteDomainById is WHERE id = $1 only. After [#3251](https://github.com/domits1/Domits/issues/3251) a tenant-less row can be taken over between the read and the delete, so this could delete another site's claim. AND site_id = $2 would match the scoping in [#3251](https://github.com/domits1/Domits/issues/3251).
If markDomainRemoving returns null, startRemoval still disables the tenant and then calls deleteCustomDomainRecord with record: null.
"Remove domain" has no confirmation, and for a live domain one click takes the site offline.
After removal the host's CNAME still points at the shared routing endpoint. Given the takeover discussion on [#3251](https://github.com/domits1/Domits/issues/3251), the removed message should tell the host to delete that record.
Please make sure the API Gateway DELETE method and the cloudfront:DeleteDistributionTenant permission are tracked, so they don't get missed after merge.

PR #3255, opened 2026-09-16
Host payout screen adjustment

BeMounir:
## Close or Relate the Issue
- Closes #

- Related Issue: #

## Proposed Changes
Description:

Redesigned the host Finance screen to match the provided wireframes across onboarding, Stripe connection, go-live, and live payout states.

Added:
- Correct onboarding order: List your property → Connect Stripe → Go live
- Real listing and Stripe account status handling
- Balance, payout, transaction, payout settings, and help sections
- Working transaction filters for All, Payments, Payouts, and Refunds
- CSV export for filtered transactions
- Expandable upcoming payouts
- Functional payout settings navigation
- Configurable finance demo data through `.env.local`
- Responsive styling for the finance dashboard

afbeelding
afbeelding

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
- None

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [x] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [ ] PR title is descriptive and includes issue number
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [ ] Jest tests are included
  - [ ] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


karimh01 (changes requested):
Three things to fix first.

**Demo mode works on production and the Connect button is not guarded**

isFinanceDemoMode (financeDemoData.js:58-67) also accepts ?financeDemo=true and localStorage, so it works on any build. In demo mode accountId stays null, so the onboarding says Stripe is not connected. handleStripeAction (refreshFunctions.js:294-318) has no demo check, so pressing Connect Stripe really calls createStripeAccount() and sends the host into Stripe onboarding, even if they already have an account.

Gate the flag on a non-production build, and refuse handleStripeAction in demo mode.

**The Refunds tab is always empty with real data**

Categories are guessed in the frontend (HostFinanceTab.js:111-138) from description and status. But the backend mapping (stripePayoutsService.js:65-83) has no description field, its status is Stripe's charge status, and it never returns refunds since charges come from transfers. So a refunded booking stays under Payments at its original amount. Your demo data uses shapes the backend never produces, which is why it looks right in demo. The category should come from the backend using Stripe's refunded and amount_refunded.

**refreshFunctions still does not read FINANCE_DEMO_DATA.account**

The demo block (lines 58-65) only fills charges, payouts, balance and schedule, and the account loader returns early on line 76. So isConnected and isLive stay false and Ready to withdraw, the steps card and the payout settings never show in demo mode.

**Smaller**

- "arriving soon" and "processing" both print processingAmount (237-238 and 297, 302), so the same number shows under two labels and twice in the summary.
- Sonar has 11 new issues: five nested ternaries in HostFinanceTab.js (263, 269, 292, 297, 302), a duplicate .finance-steps selector in the scss (1087 and 215), and five low contrast warnings (1129, 1174, 1222, 1350, 1487).
- The CSV Date column is the backend display string, 06 May 2026, so spreadsheets will not sort it as a date.
- The payouts endpoint mixes forecast rows into the list (stripePayoutsService.js:141-149, no id, status incoming charge - pending). Those show as Upcoming payout transactions on screen and in the CSV, but they are projections.

EnesVeli (changes requested):
I agree with Karim's review. I checked it against the code and found three more things.

Clarification on Karim's third point: in demo mode the account loader (refreshFunctions.js:75-89) still loads the host's real Stripe account. It only returns early when the host has no account. So a host with a real account sees demo balances on their real account. Demo mode should set the account from FINANCE_DEMO_DATA.account and skip the real call.

Blocking

Stripe account issues are no longer shown. The previous screen showed "Action required: Stripe account issue detected" when chargesEnabled or payoutsEnabled was false. HostFinanceTab.js no longer reads those fields (78-96) and shows "Bank account connected and ready for payouts" based only on accountId && onboardingComplete (101, 404-406). A host whose payouts Stripe has disabled is told everything is fine.
Real balances are hidden when no listing is ACTIVE or when the listings request fails. showFinancialData depends on isLive, which requires an ACTIVE listing (105-106, 149-151). A connected host who pauses their listing sees €0 and no payouts. If fetchHostOwnedListings fails, the catch sets hasProperty: false (158), so an existing host is sent to "List your property" and sees no financial data. Balances and payouts should show whenever Stripe is connected, whatever the listing status, and a failed listings request should show an error instead of the onboarding state.
"Save schedule" writes to real Stripe in demo mode. handlePayoutSchedule (refreshFunctions.js:220-247) has no demo check, and demo mode sets the interval to "daily" (62). A connected host in demo mode who clicks Save overwrites their real payout schedule. It needs the same guard as handleStripeAction.

Non-blocking

There are no tests for the derived state (isConnected, isLive, showFinancialData, transaction type). That's business logic, and a test would have caught point 2.
Nested ternaries are still in HostFinanceTab.js:269-279.
The FAQs are hardcoded (51-55) but getFaqs() still runs on every load (refreshFunctions.js:145-156).
"Withdraw Funds" opens the Stripe account link. It doesn't withdraw anything.

EnesVeli (dismissed):
I trust you bro

karimh01 (changes requested):
Nice, everything from the first round is fixed. Demo mode is off in production and can't touch Stripe anymore, refunds come from the backend now, balances stay visible without an active listing, and the Stripe account warnings are back. Good that the derived state has its own tests.

"Arriving soon" and "Last payout" use the same payout row. Both read recentPayouts[1] (HostFinanceTab.js:274 and 335). With real data payouts[0] is the forecast, so payouts[1] is either a projection or the last paid payout. Either the Last payout date is in the future, or paid-out money shows as arriving soon. Take Last payout from the first row with isProjected false, and arriving soon from the projected rows after the forecast.

Two small things, not blocking:

A partial refund hides the whole payment. stripePayoutsService.js:58 marks a charge as refunds as soon as amount_refunded > 0, so a 50 refund on a 670 booking only shows -50. Fine for a follow-up.

The PR title has no issue number and Closes # is empty.

Request changes, just for the arriving soon / last payout row.

karimh01 (dismissed):
Checked the fix.

Last payout now takes the first row with isProjected false and arriving soon sums the projected rows after the forecast, both with a test. That fixes the mixed-up numbers.

One small thing, not blocking:

getArrivingSoonAmount skips payouts[0] with .slice(1), assuming it is the forecast. The backend only adds a forecast row when it has one, so without a forecast the first real projection is left out of arriving soon. Filtering on the forecast itself (for example id === null && status !== "incoming charge - pending") instead of position would be safer.

Approve.

Nishchal-code (dismissed):

EnesVeli (approved):

Mehedihasanb (approved):

Issue #512, opened 2024-04-09
Website: Guest dashboard (Bookings tab)


stefanhopman1:
### Feature summary 
The  guest dashboard bookings tab tells a guest all relevant information related to their bookings. Compare existing booking platforms to create a good design / implementation. 

----

<details>

<summary> <strong> V1 </strong> </summary>

- [ ] #2813 
- [x] Backend
- [ ] Route: https://www.domits.com/guestdashboard/bookings

</details>

V2
- [ ] #2816 
- [ ] Backend
- [ ] Testing E2E & Unit

Overview
- [ ] General overview of different bookings
- [ ] Current bookings
- [ ] Upcoming bookings
- [ ] Previous bookings
- [ ] When clicked, some information about booking

PR #3256, opened 2026-09-16
Feature/512 reservation details fixes

Rakib730066:
## Close or Relate the Issue
- Closes #

- Related Issue: #512 

## Proposed Changes
Description:
This PR covers reservation-details work on #512 (Bookings tab):

1. Fixed check-in/out instructions (previously hardcoded empty) — now shows 
   the property's check-in/out time window as a fallback.
2. Added an Amenities section, reusing the existing amenity catalog resolver.
3. Implemented Special Instructions end-to-end (host-authored notes like 
   parking/WiFi) — activated previously-stubbed backend functions, wired 
   through the controller, added frontend display.
4. Implemented Special Requests end-to-end (guest-editable notes like late 
   check-in) — new repository method, service with guest-ownership auth 
   check, controller PATCH action, frontend API call, new UI section with 
   textarea/save button, styled to match the page.
5. Addressed 5 Copilot review comments: fixed database error handling to 
   only swallow the "not-yet-migrated" case (not real DB failures), fixed 
   search filtering to cover all booking sections (not just the main tab), 
   fixed the error state to no longer show alongside a misleading "no 
   bookings" message, and fixed always-blank tab counts.

Important: two database migrations are needed before Special Instructions 
and Special Requests show real data — the `property_custom_rules` table 
and a `special_request` column on `Booking`. Both features are safely 
wrapped so they won't break anything before those migrations land; they'll 
just keep showing empty states until then.

## Change Type
- [x] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
- src/features/example.xx
- N/A — no refactoring, only new code added and bug fixes.

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [x] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [ ] Conventions
- [ ] Testing
  - [x] Code tested locally
  - [ ] Jest tests are included
  - [ ] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


Mehedihasanb (commented):
Good progress addressing the earlier Copilot round — most of it checks out. Two things I don't think are actually resolved yet, plus one new one:

**Please confirm `Property_CustomRule` is registered in `Tables.js`.** The thread on `getCustomRules` shows "Resolved as Addressed," but `Tables.js` isn't part of this diff so I can't see the fix directly. This one matters beyond the new feature — `getCustomRules` runs inside `getFullPropertyAttributes`, so if the entity still isn't registered, every reservation/listing-detail request that reaches this path throws `EntityMetadataNotFoundError`, not just "shows an empty state."

**`normalizeCustomRules` still drops the `enabled` flag.** Copilot's suggested changeset for this hasn't been applied — the current code still only maps `category` and `rule_text`, no `enabled`. Concretely: a host disables a custom rule, saves any other overview change, the flag gets stripped, the repository treats the omission as enabled, and `SpecialInstructionsSection` (`enabled !== false`) shows the "disabled" rule to guests. One-line fix (`enabled: rule?.enabled !== false`) per Copilot's suggestion.

**New one, on the write side of special requests:** `updateBookingSpecialRequest`'s not-yet-migrated fallback (`42P01`/`42703`) returns `getBookingById(id)` wrapped in the exact same `{ response, statusCode: 200 }` shape as a real successful update. `SpecialRequestsSection` only checks that the fetch resolved, so until the `special_request` migration lands, every guest who saves a request gets a "Special request updated." success toast and an optimistic UI update, and nothing is actually persisted — a refresh silently loses it with no error ever surfaced. That's worse than the read-side's honest empty state, since it's false positive feedback on something the guest actively did. Worth either a distinct "not available yet" response the frontend can detect, or gating the section behind the migration status until it lands.

Smaller, non-blocking: the inquiry-booking search gap Copilot flagged (property prefetch not covering `inquiryBookings`) doesn't look fixed in what's here. And the check-in fallback string would render a literal "undefined" if `till` is ever missing while `from` is present — cheap to guard against both.

Nishchal-code (changes requested):
Please change before merge:

Add special_request to the Booking model. I couldn't find it there, so the save will likely never work. Please confirm this on the PR branch.

Narrow the catch in updateBookingSpecialRequest. It catches every TypeORMError and hides real bugs. Only catch the "column does not exist" error (42703), and remove this fallback after the migration.

Validate the special request on the backend. Check that it is a string and set a max length (for example 500). Add maxLength to the textarea too.

Run Prettier. The indentation is broken in propertyService.js, GuestBooking.js and ReservationDetails.jsx. There are also extra blank lines in propertyController.js and missing newlines at the end of some files.

Add tests for the new backend logic. updateSpecialRequest needs three cases: not the guest → Forbidden, missing booking → 404, success. The search and tab filter needs one test.

Nishchal-code (dismissed):

PR #3258, opened 2026-09-16
#3106 Log the CloudFront error and record the AWS setup

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3106

## Proposed Changes
Description:
The domain sync fails on acceptance with tenant_create_failed and there is no way to tell why. Four catch blocks in websiteCustomDomainService turn a CloudFront error into one of our own codes and throw the original away: createOrAdoptTenant, the sync, and both removal steps. The Lambda writes no log line at all, so CloudWatch only shows that the function ran.

Added one helper in the same shape the rest of PropertyHandler already uses, a plain console.error with a short prefix. It logs the operation, the domain and the site id plus the error object, which carries the AWS error name, message and request id. Tenant ids and certificate ARNs stay out of the log.

I ran into this today: creating a tenant by hand with the CLI works, the same thing through the service fails, and there is nothing to go on.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
Nothing refactored.

## Evidence
The existing tests already cover all four failure paths, so the new lines show up in the test output with the real error attached.

## Responsiveness
No UI.

## Npm Packages

## Checklist
- [x] Conventions
  - [x] No config files
  - [x] No global styling
  - [x] No console.log
  - [x] No commented-out code
- [x] Testing
  - [x] Code tested locally
  - [ ] Jest tests are included
  - [x] Jest tests are passing, 72 in the two PropertyHandler domain test files

## Keep or delete my branch
- [ ] Keep my branch


EnesVeli (changes requested):
thanks, this fills a real gap. Right now a failed domain call leaves nothing in CloudWatch, and the helper uses the same [CustomDomain] shape as the logs you added in [#3287](https://github.com/domits1/Domits/issues/3287). I also checked the IAM list in the design pack against cloudFrontTenantRepository.js:3-9, and it matches the seven commands exactly. The current doc on acceptance leaves out DeleteDistributionTenant, so that fix is needed too.

Needs a change before merge

The branch conflicts with acceptance in websiteCustomDomainService.js since [#3287](https://github.com/domits1/Domits/issues/3287) landed. [#3287](https://github.com/domits1/Domits/issues/3287) moved the createOrAdoptTenant catch into mapTenantCreateFailure (acceptance :209, :217), and that helper throws TENANT_CREATE_FAILED without logging. If the conflict is resolved by keeping acceptance's side, the tenant_create_failed log, the one this PR started from, disappears without anyone noticing. Could you rebase and put logCloudFrontFailure({ operation: "CreateDistributionTenant", ... }) directly before the throw in mapTenantCreateFailure? That also covers the retry path in retryCreateTenant.

Smaller points, not blocking

In the sync catch (websiteCustomDomainService.js:370 on this branch) the log runs before the isForeignTenant check, so a tenant_not_owned refusal gets logged as a CloudFront failure. Logging only when !isForeignTenant would keep the log to real CloudFront errors.
Could the PR title follow the conventional-commit format, like the commits do? For example fix(domains): log CloudFront errors and document the AWS setup (#3106).
Heads-up: this PR and #3262 both touch the same catch blocks and conflict with each other, so whichever merges second needs a rebase.

EnesVeli (approved):
@karimh01 thanks, verified. The log now sits in mapTenantCreateFailure right before the TENANT_CREATE_FAILED throw (websiteCustomDomainService.js:254), so it only fires on a real failure and also covers the retry path. The branch merges cleanly with acceptance. The red web build is the useCompanyLogoUpload failure inherited from acceptance and is being fixed in [#3297](https://github.com/domits1/Domits/issues/3297). Two optional leftovers, no need to hold the merge for them: the sync catch (:499) still logs a tenant_not_owned refusal as a CloudFront failure, and the title could follow the conventional-commit format. Approving.

Nishchal-code (approved):

PR #3261, opened 2026-09-17
feat(personal-data): add password management to Personal Data settings

Nishchal-code:
## Close or Relate the Issue
- Closes #
- Related Issue: #2560

## Proposed Changes
Description:
Adds password management to the Personal Data settings page (host and guest dashboards), under Preferences / Authentication:
- The password is shown as a masked placeholder (hidden characters), never the real value.
- A "Change Password" button expands an inline current/new/confirm-password form in place — no popup/modal.
- Each password field has its own eye icon to show/hide what was typed, independent of the other fields.
- Client-side validation covers empty fields, minimum length, and new-password/confirmation mismatch before calling Cognito.
- Backed by a new `usePasswordChange` hook that wraps Amplify's `Auth.changePassword`, wired into `useSettingsData` so both host and guest settings pick it up automatically.
- Added matching copy across en/nl/de/es content files.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
N/A — no files refactored without behavior change.

## Evidence
- [] Screenshots or screen recording added for UI changes
- [] Before/after images added when relevant
- [] Images clearly show the implemented change
- [] Image quality is readable (no cropped or blurry evidence)

## Responsiveness
- [x] UI checked on mobile
- [ ] UI checked on tablet
- [x] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

N/A — no dependency changes, uses existing aws-amplify Auth API.

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


Mehedihasanb (commented):
The Quality Gate is still failing on Security Rating, so this isn't mergeable yet regardless of approvals — flagging where I think it's still coming from.

The last commit scoped the `S2068` (hard-coded credentials) suppression to `frontend/web/src/content/**/*.json` only:

```
sonar.issue.ignore.multicriteria.e3.ruleKey=json:S2068
sonar.issue.ignore.multicriteria.e3.resourceKey=frontend/web/src/content/**/*.json
```

That fixes the i18n false positives but doesn't touch the JS side. `usePasswordChange.test.js` has literal password-shaped strings (`"oldPass1!"`, `"newPass1!"`, `"wrongPass1!"`, `"differentPass1!"`) right next to variables named `currentPassword`/`newPassword` — the same heuristic that flagged the JSON files. Can you check the SonarCloud PR page for the exact still-open finding? If it's this, the fix is the same pattern already used for the JSON files: a scoped `S2068` suppression for the test file with a comment noting these are fixtures, not real secrets.

Two things worth addressing in the code itself, separate from the CI gate:

1. The validation messages in `usePasswordChange.js` ("Please fill in all password fields.", the length check, the mismatch check, and the generic failure fallback) are hardcoded English, while every other label on this form got full en/nl/de/es copy. A non-English user filling this in wrong sees English error text.

2. `setPasswordError(error?.message || ...)` passes Cognito's raw exception text straight to the UI — untranslated and unfiltered, whatever AWS happens to word it as (wrong current password, a policy violation, rate limiting, etc.). Worth mapping the realistic Cognito error codes here to the same translated messages as the rest of the form.

Smaller, non-blocking: after a successful change the button still says "Cancel" (nothing left to cancel at that point — "Close"/"Done" would read better). And there's no `<form onSubmit>`, so Enter in the password fields won't submit — assuming that's intentional, just flagging it.

Mehedihasanb (approved):

Rakib730066 (approved):
frontend/web/src/content/**/*.json won't match content/en.json directly, since **/ requires at least one subdirectory in between. That's why the Password warnings are still showing on en.json/nl.json in this diff even after the suppression was added. Should be frontend/web/src/content/*.json instead (no **).

PR #3262, opened 2026-09-17
#3106 Let the host make a custom domain the main address

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3106

## Proposed Changes
Description:

A host could connect his own domain and it would serve his site, but Domits kept showing the long `.direct.domits.com` address everywhere: the "Open live site" button, the notice after publishing, and anywhere a link to the site is built. This adds a "Make main address" button on a live custom domain, and a "Main address" badge on whichever domain holds the flag.

**Who owns what.** The row that gets promoted is always the site's stored custom domain, never one chosen by the request value. A mismatch answers `domain_not_found`, the same as remove. Only an `ACTIVE` custom domain can be promoted.

**Exactly one main address.** The move is a single `UPDATE` scoped to `site_id` that sets `is_primary = (id = target)` on every row of the site, with a subquery that re-checks the target is still `CUSTOM` and `ACTIVE` at write time. A concurrent removal or status change makes the statement change zero rows rather than leaving two primaries or none.

The flag is handed back to the fallback in the same request that moves the custom domain off `ACTIVE`: removal before the tenant is disabled, a sync landing on `FAILED` or `DISABLED`, and the row delete. So the main address is never a dead row.

**Two bugs found on the way.** The fallback upsert on publish did `is_primary = EXCLUDED.is_primary`, so every republish reset the flag to the fallback. With promotion on top that would have produced two primaries. It now keeps the stored flag on conflict. And the KPI join for fallback reachability filtered on `is_primary = TRUE`, so after a promotion the fallback subdomain would have counted as unreachable.

**Concurrency.** Zero changed rows is also what a second, concurrent promote of the same domain sees. The service rereads the stored row in that case and answers the current list when it is still `ACTIVE` and primary, instead of refusing something that already worked. Three codes where there was one: `domain_not_active` for a row that is genuinely not live, `primary_changed` (409) for a row that is live but lost the flag to a concurrent hand-back, and `domains_unavailable` (500) for the list read failing after the mutation already committed. Each has its own host copy, because all three used to show the same line saying the domain was not live.

**API compatibility.** Verify and remove now answer `{ siteId, domains }` like promote, instead of `{ domain }`. A sync or removal can hand the flag back, and the panel has to show the fallback as the main address again without a reload. Re-fetching after each action would have run a second CloudFront sync per click. The `{ domain: null }` removed answer is gone; the custom entry is simply absent from the list. Both endpoints keep their method, path and CORS, so no gateway change is needed for that. Nothing else in the repo reads those responses.

**Note on the diff:** the commits touching `directBookingWebsiteRepositories.returning.test.js` and the `runStatement` helper come from #3106 Read UPDATE and DELETE RETURNING results correctly, merged in because promotion depends on it. Once that PR lands they disappear from this diff.

**Console work after merge:** the new route `/property/website/domains/primary` needs an API Gateway resource, POST as Lambda proxy, OPTIONS with the method in `Access-Control-Allow-Methods`, a Lambda invoke permission for that path, and a stage deploy.

**Known and tracked separately:** a sync or provisioning request that read an outdated row can still overwrite newer state, including after a promotion. That needs a guard on those writes, and refusing a stale provisioning write without disabling the tenant of the request that won needs its own handling. Both are in #NUMMER.

**Not needed:** repairing sites broken by an earlier partial hand-back. Promotion never shipped, so no production site can be in that state.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
The four domain endpoints now build their response through one shared controller helper, `buildWebsiteDomainsResponse`, instead of each shaping its own body. Same output for GET and promote, new shape for verify and remove as described above.

## Evidence
The main address flow itself is covered by tests; the button and badge need the new route on the gateway before they work end to end on acceptance.

Verified separately against the real database that the two new statements run on Aurora DSQL: `IS DISTINCT FROM`, an `EXISTS` subquery inside an `UPDATE`, and the `INSERT ... AS existing` alias are all supported. Ran them as a dry run on a row that was already primary, so nothing changed.

## Responsiveness
The button and badge sit in the existing domain panel and follow its layout.

- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [x] UI checked on desktop

## Npm Packages

- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Known limitations
Two things I found while reviewing this and deliberately left out, both older than this change and both getting their own issue:

- `ensureDomain` overwrites `site_id` on conflict, so two concurrent claims can transfer a domain row between sites. Same root as the ownership race in #3251.
- Status changes and deletion commit separately from the flag reassignment. Without a transaction, a failure in between can leave a site with no main address.

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing, 137 backend and 284 frontend in the domain suites

## Keep or delete my branch
- [ ] Keep my branch


karimh01:
API Gateway is set up for the new route: resource primary under domains, POST as Lambda proxy, OPTIONS with CORS, the invoke permission and a stage deploy. Verified with an OPTIONS call, which answers 200 with the right headers.

So this works on acceptance as soon as it merges. Right now a POST to that path falls through to the create handler, because the Lambda runs the code from acceptance where the route does not exist yet.

EnesVeli (commented):
nice work. Handing the flag back on every path that takes a custom domain off ACTIVE is thorough, the promote UPDATE keeps "exactly one main address" in a single statement, and the three separate error codes give the host honest copy. I checked the rebase ahead of time: only one test hunk and one doc bullet conflict with acceptance, and the service, repository and controller merge cleanly. The merged ensureDomain keeps both your is_primary = existing.is_primary and #3287's WHERE existing.site_id = EXCLUDED.site_id. With the test hunk resolved as keep-both, your 5 domain suites pass (92/92), and the full backend suite shows only the suites that need a DB or AWS credentials. I'll approve once the rebase is pushed.

karimh01:
**Changed after review:** a repeated connect from the same site no longer resets a live domain or drops its tenant, it returns the existing row. The status change and the hand-back of the primary flag now happen in one transaction, retried once on a DSQL commit conflict. A failed rollback discards the connection instead of returning it to the pool. The hand-back tests now run against a storing fake instead of a mock that echoed its input.

EnesVeli (approved):
@karimh01 thanks, verified. The rebase is clean, and the new transaction code holds up on DSQL: startTransaction() sends a plain START TRANSACTION, the retry re-runs the whole unit on 40001/OC001, and I checked in the TypeORM and pg-pool source that release() doesn't roll back and that releasePostgresConnection(err) removes the client from the pool. The claim keeps #3270's per-site 23505 mapping because ON CONFLICT (domain) only covers the domain index, and a foreign row still answers domain_taken. All 11 domain suites pass locally (212 tests). Approving.

Could you update the description before merging? #NUMMER is still a placeholder, the #3259 note and both "Known limitations" no longer apply (the first was fixed in [#3287](https://github.com/domits1/Domits/issues/3287), the second by this PR), "Not needed: repairing sites…" contradicts the unconditional restore in b9747f4c, and the size is now +2268, so "Huge change".

Merge order: this PR and #3258 still conflict in websiteCustomDomainService.js. Could you merge this one first and then rebase [#3258](https://github.com/domits1/Domits/issues/3258) on top (it's only the four log lines), and text me so I can look at that resolution before it goes in?

karimh01:
Description updated: the placeholder is filled in, both stale notes are gone, the "not needed" line is out since the restore now runs unconditionally, and the size is set to Huge. Looking for a second approval, then I'll merge this and rebase #3258 on top and send it your way.

BeMounir (approved):
Looks good overall, just a few notes. the SQL concurrency test doesn’t actually use two separate transactions/sessions, promotion doesn’t retry 40001/OC001 conflicts, and `getPrimaryLiveDomainBySiteId()` is a bit misleading since it returns the fallback domain. but overall it looks good

Nishchal-code (approved):
Approved. The promote query, the transactions and the race-case tests look solid.

PR #3264, opened 2026-09-17
fix(bookings): treat hosts with no properties as an empty result, not an error

Nishchal-code:
## Close or Relate the Issue
- Closes #
- Related Issue: #

## Proposed Changes
Description:
Hosts with zero active properties were seeing a "Failed to load reservations" error on the Reservations page instead of an empty state. Root cause: `getPropertiesFromHostId` threw a `NotFoundException` (404) whenever the property-lookup service reported no properties for a host, even though that's a normal, valid state (e.g. a new host with no listings yet) — not a failure. The exception bubbled up through `readByHostId` → `bookingService.read()` → the API response, and the frontend's generic error toast fired.

Fix: `getPropertiesFromHostId` now returns an empty array when the host genuinely has no properties, and only throws when the response is malformed/unrecognized (a real failure). The frontend already had the correct empty-state handling (`setBookings([])` → "No reservations yet") — it just needed to stop being told this was an error.

Verified live by deploying the fix to the `General-Bookings-CRUD-Bookings-develop` Lambda and confirming the Reservations page now shows "No reservations yet" with no error toast for a host with no properties.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
N/A — no files refactored without behavior change.

## Evidence
- [] Screenshots or screen recording added for UI changes
- [] Before/after images added when relevant
- [] Images clearly show the implemented change
- [] Image quality is readable (no cropped or blurry evidence)

## Responsiveness
N/A — backend-only fix, no UI changed.

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

N/A — no dependency changes.

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


Rakib730066 (approved):

karimh01 (approved):
Small and correct. A host with no properties now gets an empty list with a 200 instead of a 404, so the reservations page shows "No reservations yet" instead of the error toast.

I checked every place that reads this. The reservations page, the dashboard, the calendar and the reservation details all handle an empty list fine, nothing continues where it should not. And only the exact "No property found." signal becomes empty. A network failure or a database error on the Bookings side still answers 500, so nothing real gets swallowed.

Two things outside this PR, worth a follow-up:

When PropertyHandler itself fails with a 5xx, the body is "Something went wrong, please contact support.", which ends up as a 404 "User has no active properties." The host still sees the error toast, but the status and message are wrong. That was already the case before your change.

The two existing mapping tests in lambdaRepository.test.js call the real Database.getInstance() for the rules query without a mock, so they only pass with AWS credentials. That also predates this PR, but it means this test file is not a clean CI signal until the database is mocked.

PR #3265, opened 2026-09-17
fix(inbox): use access token for listing details (#2291)

elcoleta:
## Close or Relate the Issue
- Closes #

- Related Issue: #2291

## Proposed Changes
Description:

Fixes the host inbox listing panel failing to load listing details.

`ListingPanel` was using the ID token from `AuthContext` when calling
`GET /property/hostDashboard/single`, but this endpoint requires the Cognito access token.

This change:
- Uses the real access token when loading host listing details
- Leaves the existing ID-token consumers unchanged
- Keeps the guest listing path unchanged
- Adds regression tests for both host and guest behavior

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring

N/A - no refactoring.

## Evidence
N/A - no visual or layout changes.

## Responsiveness
N/A - no responsive or layout changes.

- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
N/A - no packages added, removed, or updated.

- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

Testing completed:
- ListingPanel tests: 2/2 passed
- Relevant messaging/frontend tests: 57/57 passed
- ESLint: 0 errors, 0 warnings
- `git diff --check`: clean

## Keep or delete my branch
- [ ] Keep my branch


karimh01 (approved):
Correct fix, and dropping accessToken from the deps is right since the component no longer reads it.

Checked Copilot's race and it is not reachable. The panel cannot fire before the session resolves, because propertyId only gets a value after auth is established: the navigation path sets it only when userId is truthy, which comes from Auth.currentAuthenticatedUser, and the thread path needs a contact from useFetchContacts, which awaits getIdToken first. Both refresh the session and write the tokens to localStorage before resolving, so by the time the effect runs with a non-null propertyId the access token is there. And getAccessToken is synchronous, reading the Amplify-persisted token from localStorage, so there is no "not yet ready" state that resolves later.

One thing worth knowing: on acceptance the effect ran twice for a host, once with Authorization: null (fetch sends the string "null") and then again with the ID token. Both got 401, so the host saw the error either way. Your change makes it one request with the right token.

Two small things, neither blocking. The AuthContext mock at lines 12-14 of the test is dead now that the component does not import it. And neither test covers the failure branch at ListingPanel.jsx:41-46.

The endpoint does need an access token, confirmed: hostDashboardHandlers.single goes through authManager.getAuthorizedUser to cognitoRepository.getUserByAccessToken, which is GetUserCommand({ AccessToken }), and Cognito GetUser only accepts access tokens. Guest path is genuinely unchanged, needsAuth stays false and the backend does no caller authentication there.

EnesVeli (approved):

Issue #3266, opened 2026-09-17
[Feature]: Attach the Cognito authorizer to the Channex and Holidu API routes.


EnesVeli:
## Feature Summary
Attach the existing Cognito User Pool authorizer in API Gateway to the channel-management routes (`/integrations/channex/*` and `/integrations/holidu/*`) on the UnifiedMessaging API, so requests reach the Lambda with the caller's Cognito claims.

---

## Problem Statement
Since #<PR number>, the shared channel-management handler takes the user from the API Gateway Cognito authorizer claims (`requestContext.authorizer.claims.sub`) and returns 401 when they are missing. Channex `connect` and `disconnect` are also restricted to allowlisted users.

On the UnifiedMessaging API (`54s3llwby8`), these routes are currently served only by the catch-all `/{proxy+}` (`ANY`, `authorizationType: NONE`), so no claims reach the Lambda. After #<PR number> is deployed, every request from the Channex certification admin panel returns 401.

Affected: the Channex certification admin panel (allowlisted internal users). No host-facing screen and no other Lambda calls these routes.

---

## Desired Outcome
The Channex admin panel works again for allowlisted users, with the user identity coming from the Cognito token, and no other route on the API changes behaviour.

---

## Target Users
- [ ] Guests / Customers
- [ ] Hosts / Property Managers / Enterprise
- [x] Admin / Internal users
- [x] Developers
- [ ] Other: _______

---

## Business & User Impact
- **Impact level:**
  - [x] High (unblocks key use cases)
- **User reach:**
  - [x] Few users

---

## Proposed Solution (Optional)
Current state (checked with `aws apigateway get-authorizers` / `get-resources`):
- An authorizer already exists: `DomitsCognitoAuthorizer` (id `wgzgpr`, type `COGNITO_USER_POOLS`, user pool `eu-north-1_mPxNhvSFX`, identity source `Authorization`). `/threads` GET and `/threads/{id}/read` POST already use it. Reuse it.
- There are no `/integrations` resources. Channex, Holidu, WhatsApp and `GET /integrations` all go through `/{proxy+}`.

Steps:
1. Check that the UnifiedMessaging Lambda permission allows invocation from new resources (`aws lambda get-policy --function-name UnifiedMessaging`). If it is scoped to `/{proxy+}` only, add a permission for the new paths.
2. Create resources `/integrations`, `/integrations/channex/{proxy+}` and `/integrations/holidu/{proxy+}`.
3. On both `{proxy+}` resources:
[...]

PR #3267, opened 2026-09-17
fix(channel-management): take the Channex user from the Cognito token (#440)

EnesVeli:
## Close or Relate the Issue
- Closes #

- Related Issue: #440 #3266 

## Proposed Changes
Description:

The channel-management routes (`/integrations/channex/*` and `/integrations/holidu/*`) now identify the user from the API Gateway Cognito authorizer claims instead of from the request.

**Backend (shared channel-management handler)**
- New helper `channelAuthenticatedUser.js` reads `sub` from `requestContext.authorizer.claims` (or `authorizer.jwt.claims`).
- The handler requires those claims on every user-facing channel route and returns `401` without them.
- Before routing, `userId` in the query string and in a JSON body is replaced with the token user, so controllers no longer act on a caller-supplied `userId`.
- The Channex certification allowlist is applied to the token user, and it now also applies when no `userId` is sent.
- `POST /integrations/channex/connect` and `/disconnect` are added to `protectedChannexCertificationAdminRoutes`, so only allowlisted users can use them.
- `booking-availability/sync` and `calendar-change/sync` are unchanged. They are called Lambda-to-Lambda and stay authorized by `x-domits-internal-token` in the controller.

The check lives in the shared handler rather than in each controller method, so every channel route (and any route added later) goes through one place. The handler already made the allowlist decision there before this change.

**Frontend**
- `requestChannex` in `channexApi.js` sends `Authorization: Bearer <Cognito ID token>`, the same format the inbox `/threads/{id}/read` route uses behind its Cognito authorizer. If no token can be read, the request is not sent and the error shows in the panel.

**Deployment note**
The Channex and Holidu routes are currently served through the API Gateway catch-all `/{proxy+}`, which has no authorizer, so no claims reach the Lambda. Until the authorizer is attached (tracked in #<API Gateway authorizer issue>), these routes return `401`. The only caller is the Channex certification admin panel. No host-facing screen and no other Lambda calls them.

Entry diagnostic logs for `sync/restrictions` and `sync/full` still log the `userId` from the query string. Left unchanged here to keep the diff focused.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
-none

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [x] Conventions
- [x] Testing
  - [] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


karimh01 (approved):
Correct fix, and putting the check in the shared handler rather than in each controller method is the right call.

Checked the userId replacement and it is complete. Every controller method identifies the caller through either the query string or the top-level body.userId, nothing reads pathParameters or headers for identity, and withAuthenticatedUserId overwrites both. The shapes you do not rewrite fail closed rather than open: invalid JSON, an array or a primitive all end at the 400 for a missing userId. Nested credentials.userId is never read as an identity.

WhatsApp is untouched, confirmed. isChannelHttpPath matches only channex and holidu, so a WhatsApp path falls through to UnifiedMessaging's own route table.

The allowlist does not lock anyone out who could get in before. The only connect UI is the diagnostics panel, which hides itself unless admin-access answers allowed, and the nav link is gated the same way. So the reachable set is unchanged and only the anonymous path is gone.

The internal-token routes are sound. Both refuse with 403 when the env var is missing, the skip only applies with TEST=true, and TEST is not set on any deployed Lambda. I probed one from outside and got 403.

**One thing for the deployment note, and it matters.** The messaging API has no /integrations resources at all. Everything reaches the Lambda through /{proxy+} with authorization NONE, and that same catch-all carries the WhatsApp webhook that Meta calls unauthenticated, plus the WhatsApp connect calls that send no token. So the authorizer cannot go on /{proxy+}, it has to go on dedicated /integrations/channex/{proxy+} resources. Worth writing into the note so whoever attaches it does not take WhatsApp down.

Also worth knowing for whoever picks up the follow-up: the certification access check runs for every host on dashboard load through Pages2.js, not only on the admin page. Until the authorizer is attached that gets a 401 every time, which is caught and hides the link, so no visible breakage. Just noisy.

Small, not blocking: the token comparison is a plain !== rather than constant time. Theoretical timing channel on a 64-character token, not worth changing.

elcoleta (approved):
looks good! Nice work

Issue #3263, opened 2026-09-17
Web: OTA Booking Engine Domits.com: Property Details: Price-Saving USP (~€XXX less than other sites)


stefanhopman1:
## 🚀 Feature Summary

Display a **price-saving USP directly below the property price** on the Domits.com property details page.

The message communicates to guests that they can save money by booking through Domits compared with higher-commission platforms such as Airbnb.

Example:

> **~€XXX less than other sites**
> *Save more by booking directly on Domits.*

The saving should be calculated dynamically based on the booking price and the applicable commission difference.

Example:
<img width="199" height="85" alt="Image" src="https://github.com/user-attachments/assets/b0af3ee9-201b-4abe-a529-f2e1ca97635b" />

### Commission Example

Assuming:

* Airbnb guest-facing/host commission impact: **15.5%**
* Domits commission: **10%**
* Difference: **5.5 percentage points**

For a €1,000 booking:

* Airbnb: €1,000 × 15.5% = €155
* Domits: €1,000 × 10% = €100
* Potential difference: **€55**

Therefore, the UI could display:

> **€55 less than other sites**

The calculation must be dynamic based on the property's current price and applicable booking configuration.

---

## Problem

Guests comparing the same luxury vacation rental across different booking platforms may see similar properties and prices without immediately understanding the financial benefit of booking through Domits.

Domits needs a clear, immediately visible **price advantage** to communicate why a guest should book on Domits rather than another OTA.

Without explicitly showing the potential saving, the lower Domits commission structure is not visible to the guest and therefore does not function as a strong conversion advantage.

---

## Desired Outcome

Increase guest conversion on Domits.com by making the financial advantage of booking through Domits immediately understandable.

The guest should be able to see:

1. The current Domits price.
2. The estimated amount saved compared with the selected comparison platform/standard.
3. A concise explanation of why Domits can offer the lower price.

The message should feel like a **trustworthy price advantage**, not an aggressive competitor comparison.

---

### Primary

* Guests searching for luxury vacation rentals on Domits.com.
* Guests comparing properties across multiple OTAs.

### Secondary

* Domits hosts/property managers.
* Domits commercial and growth teams.

---

### Guest

* Immediately understands the financial benefit of booking on Domits.
* Increased perceived value.
* Greater confidence when comparing Domits with other OTAs.

### Domits

* Potentially higher booking conversion.
* Stronger OTA differentiation.
* Clearer communication of Domits' lower commission structure.
[...]

PR #3269, opened 2026-09-18
Feature/3263 price saving usp

Rakib730066:
## Close or Relate the Issue
- Closes #

- Related Issue: # 3263

## Proposed Changes
Description:  Implements the Price-Saving USP feature (#3263)

1. Backend: added a new GET endpoint /property/pricing/saving-config on 
   PropertyHandler, returning commission rate configuration 
   (domitsCommissionRate, comparisonCommissionRate, comparisonPlatformLabel) 
   as a static config object. Manually added and deployed the corresponding 
   API Gateway route (this repo has no infra-as-code, per CLAUDE.md, so 
   route wiring is a manual step — confirmed with Stefan before doing this).

2. Frontend: added a fetch function for the new config endpoint, a 
   calculator (calculatePriceSaving) implementing the spec's formula 
   (Applicable Amount × (Comparison % − Domits %)) with full edge-case 
   handling (invalid/negative totals, missing config, inverted commission 
   rates, savings rounding to zero — never shows a negative or misleading 
   saving), and a display component (PriceSavingMessage) styled as a green 
   badge matching the site's existing trust-badge conventions.

3. Wired into pricing.js, right below the total row, using the real 
   booking total.

Note: since the API Gateway route was added and deployed by me directly 
(with explicit approval), it is already live at the infra level — this PR 
only needs to merge for the Lambda code to catch up and the endpoint to 
fully work end-to-end.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
- src/features/example.xx
- N/A — no refactoring, only new code added.

## Evidence

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [x] UI checked on desktop

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [ ] Conventions
- [ ] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x
  - ] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


Nishchal-code (changes requested):
This PR has a lot of unrelated stuff in it. The description says this is just the price-saving feature, but the diff also includes special requests, custom rules/amenities, and guest dashboard search changes — none of which are mentioned in the description. Can we split this into separate PRs? It's hard to review everything together, and if one part needs to be reverted later, it'll take the others down with it.

Errors are being hidden too broadly in two places (bookingService.js and propertyService.js). Right now, any database error gets treated as "not migrated yet" and quietly ignored instead of showing a real error. That's fine for now, but once the migration is done, this could hide real bugs without anyone noticing. Can we make this more specific, or add a reminder to remove it later?

There's a hardcoded API link in the frontend (fetchPricingSavingConfig.js) instead of using whatever config setup the rest of the app uses. This means it'll break outside of production. Can we fix that to match the existing pattern?

Mehedihasanb (commented):
Good to see the three things from #3256 properly closed out here — entity registration, the `enabled` flag round-trip, and the special-request false-success toast (the `persisted` flag plus the local `savedValue` state is a clean fix for the stuck-Save-button issue too). Inquiry-booking search is fixed and now has a dedicated test.

Two things worth a look before merge:

**The error handling in `updateBookingSpecialRequest` and `getCustomRules`/`updateCustomRules` got broader than it needs to be.** Both switched from catching specific Postgres codes (`QueryFailedError` with `42P01`/`42703`) to `instanceof TypeORMError`, which covers basically anything TypeORM can throw. Now that `Property_CustomRule` is in `Tables.js`, the reason to widen it (an `EntityMetadataNotFoundError` slipping past) shouldn't apply anymore — the only realistic pre-migration failure left is a genuine `QueryFailedError` for a missing table, which the old narrow check already caught. As it stands, a real constraint violation or query bug would also get silently logged as a warning and turned into "not yet migrated" instead of surfacing as an actual error. Worth narrowing back to the specific codes now that the registration issue is fixed separately.

**`fetchPricingSavingConfig.js` hardcodes the full API Gateway URL** instead of using the shared base-URL constant the rest of the PropertyHandler-facing code uses (same pattern as `API_BOOKINGS_BASE` in `bookingAPI.js` in this same PR). That means this one call won't follow whatever environment config the rest of the app relies on, and it's a URL someone has to remember to update by hand if the domain ever changes.

Smaller: the SonarQube duplicate-selector warning on `.reservationPage` in `guestReservationDetail.scss` (line 53 and line 634) doesn't look resolved — worth merging into the one block. The optional-chaining suggestion in `calculatePriceSaving.js` is trivial either way.

The price-saving calculation itself is solid — commission-inversion guard, zero-rounding guard, and the config-fetch-failure path all check out, and the test coverage is thorough.

Rakib730066:
Closing this PR — it grew to include unrelated changes (Special Requests, 
Custom Rules, search fixes) that belong to #512, alongside the actual 
Price-Saving USP work, making it hard to review and risky to revert pieces 
independently. Reopening this as a clean, scoped PR containing only the 
USP-specific changes.

PR #3270, opened 2026-09-18
#3106 Enforce one custom domain per site in the database

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3106

## Proposed Changes
Description:

The one-custom-domain-per-site rule only lived in the service. requestCustomDomain reads the site's custom row and refuses a second one, but the check and the write are two steps, so two requests arriving together can both pass the check and both store a row. This adds a partial unique index so the database refuses it regardless of timing.

The index is on `standalone_site_domain (site_id) WHERE domain_type = 'CUSTOM'`, main schema only. The test schema never had these tables. The predicate covers every CUSTOM status including REMOVING, which matches the documented rule: a domain that is being removed still blocks a new one until removal finishes.

**Why the service change comes with it.** The sequential check still fires first and the host still sees the same 409, so nothing changes for an ordinary request. The index only decides the race. When it does, the loser's upsert raises SQLSTATE 23505 on this index, which nothing was catching, so that host got a generic 500 for something that is really "you already have a domain". It now maps to domain_limit_reached.

The match is exact on the index name, so a conflict on the other unique index (the one on `domain`) is rethrown untouched and keeps answering domain_taken. If the reread of the winning row fails, the host still gets domain_limit_reached with wording that does not name a domain, and the read failure is logged.

**Running it.** The repo does not run migrations, so this goes by hand like the booking columns in week 1. There is a runbook at `docs/internal/tools/dsql_custom_domain_index_runbook.md` with the order of operations, the completion criteria, and what to do when the async build fails.

**Two things the runbook has to verify before this mapping can be trusted.** Both need the index to exist, so they cannot be done here:

- Check A captures the real error through the same database package and pg driver the Lambda uses, and records `code`, `constraint`, `schema` and `table` as the driver reports them. The service matches on `error.constraint`, and that is a separate protocol field from the error message, so the catalog name and the message text prove nothing about it. If the captured constraint does not equal `standalone_site_domain_custom_site_unique`, the constant has to change before this is deployed.

- Check B runs the real upsert statement in two shapes and records what the losing one answers. DSQL can answer 40001 on concurrency rather than a unique violation, and if it does, the loser still gets a 500 and a follow-up is needed.

**Order of operations:** create the index, wait for it to go valid, run check A, then deploy. Deploying first is not harmful for ordinary requests, but index-first with the service later leaves violations on the old 500 path, and in either order an unverified constant gives a silent mismatch.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
Nothing refactored. The limit-reached error is now built by one helper so the sequential check and the race answer with the same message.

## Evidence
Checked the live database read-only: main.standalone_site_domain has 12 FALLBACK rows and 1 CUSTOM row, and no site holds more than one CUSTOM row, so nothing violates the index today. Confirmed against the AWS reference that DSQL supports partial unique indexes and the `CREATE UNIQUE INDEX ASYNC ... WHERE` grammar, which is the same family the existing migrations use.

Also captured the real error shape from the existing unique index on `domain` by forcing a duplicate inside a transaction and rolling it back: `code: "23505"`, `constraint: "standalone_site_domain_unique"`, unqualified, with schema and table as separate fields. That is why the service matches on the bare name. The same has to be confirmed for the new index once it exists, which is check A.

## Responsiveness
No UI changes.

## Npm Packages

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing, 75 in the two PropertyHandler domain test files

## Keep or delete my branch
- [ ] Keep my branch


Nishchal-code (approved):
Nice work, this is a solid fix and the runbook is thorough.

A few quick things:

The error mapping relies on the constraint name matching exactly but we can't confirm that matches the real database until "Check A" is run after the index is created. Can we make sure that actually happens before we trust this in prod?

There's a known gap (the 40001 case) that isn't fixed here should we open a follow-up ticket so it's not forgotten?

Since the migration is run by hand and order matters, who's doing that and when?

Code and tests look good just want the manual steps confirmed before merge.

elcoleta (approved):
The actual code change looks sound: the partial unique index enforces one CUSTOM domain per site at the DB level, the service maps the expected 23505 on that specific index to domain_limit_reached, and other unique violations are deliberately rethrown. The tests cover the race-loser mapping and fallback wording.
The runbook is also careful about the manual DSQL rollout and explicitly says not to run the whole SQL file top-to-bottom; it requires creating/validating the index, running Check A/B, then deploying the service.

EnesVeli (changes requested):
Good work. 
I'm requesting changes because of the conflict with #3287, not because of the design. The branch settings don't dismiss approvals on a new push, so the resolved version could otherwise merge without anyone seeing it.

Needs a change before merge

In requestCustomDomain the conflict is between #3287's null guard and storeCustomDomainClaim. Please keep both: call storeCustomDomainClaim(...), then keep the if (!record) throw ... DOMAIN_TAKEN right after it. Since [#3287](https://github.com/domits1/Domits/issues/3287), ensureDomain returns null when another site owns the domain (WHERE existing.site_id = EXCLUDED.site_id). Without the guard, the host gets 201 { domain: null } and a SITE_DOMAIN_REQUESTED event is recorded for a domain they didn't get. The test "answers domain_taken and changes nothing when a delayed claim finds the row now belongs to another site" (websiteCustomDomainService.test.js:145 on acceptance) covers this, so please keep acceptance's side of that test hunk too. I'll re-review after the push.

Smaller points

The runbook says check B uses "that exact statement text" (dsql_custom_domain_index_runbook.md:114), but the upserts in the .sql file (:60-70, :77-88) still have the pre-[#3287](https://github.com/domits1/Domits/issues/3287) DO UPDATE SET site_id = EXCLUDED.site_id without the WHERE. The racing rows use different domains, so the result probably still holds, but could you update the statements to match ensureDomain as it is now?
Check B case 1 autocommits and case 2 commits into main before the cleanup DELETE, so rows briefly exist in the live table. Could the runbook say this explicitly next to the cleanup step?
@Nishchal-code's three questions are still open. Could you reply on the PR with who runs the index and check A, and when, and whether a follow-up for the 40001 case gets opened now or only after check B?

karimh01:
@EnesVeli the conflict is resolved in 3bbfe1a44 and both sides are kept. The null guard sits inside storeCustomDomainClaim, after the try/catch rather than inside it, so a refused claim still throws domain_taken before SITE_DOMAIN_REQUESTED is recorded, and the 23505 catch cannot intercept it. The acceptance test at :145 is kept along with the three per-site index tests, and all nine domain suites pass, 164 tests. The two runbook points are fixed in the latest commit.

@Nishchal-code on your three questions:

Check A has run. I created the index on main this morning, confirmed it is valid and unique on site_id with the predicate domain_type = 'CUSTOM' only, and ran check A through the same database package the Lambda uses. error.constraint comes back as standalone_site_domain_custom_site_unique, exactly the constant in the service. The result is recorded in the runbook.

The 40001 case is tracked in #3293. I will close it or turn it into a fix once check B shows which error DSQL actually gives. Until then a 40001 behaves the same as today, a 500, so this PR does not make it worse.

I ran the migration myself. The index exists, so the service change can be deployed as soon as this merges.

BeMounir (approved):

EnesVeli (approved):
@karimh01 thanks, verified. The resolution keeps both sides correctly: the null guard after the try/catch in storeCustomDomainClaim throws domain_taken before SITE_DOMAIN_REQUESTED is recorded, and all four tests are there (the index tests at :145, :173, :189 and the guard test, now at :221). The check B statements match ensureDomain now, the live-table note is clear, and the check A record shows constraint matching the constant exactly. The red web build is the useCompanyLogoUpload failure inherited from acceptance and is being fixed separately. Since the index is already live, merging soon is best. Approving.

Issue #3191, opened 2026-09-08
Host: Account Settings: Rate Plans: Enterprise


stefanhopman1:
## 🚀 Feature Summary

Provide enterprise customers with a simple, transparent pricing model based on the number of active properties managed in Domits. Enterprise Rate Plan starts at 100+ properties. 

### Enterprise Pricing

**$49 per active property per month**

The customer pays only for active properties.

No complex module pricing, usage fees, or complicated rate components.

---

## Problem Statement

Enterprise customers need predictable pricing that is easy to understand and scale with.

Complex enterprise pricing creates unnecessary friction for:

* Property managers
* Finance teams
* Procurement
* Sales
* Account management

The pricing model should be simple enough that a customer can immediately calculate their monthly Domits cost.

---

## Desired Outcome

An enterprise customer should be able to answer:

> **"How much does Domits cost me?"**

in a few seconds.

The answer should simply be:

**Active Properties × $49**

Example:

```text
300 properties × $49
= $14,700 / month
```

---

## Target Users

* [ ] Guests / Customers
* [x] Hosts / Property Managers / Enterprise
* [x] Admin / Internal users
* [x] Finance
* [x] Sales
* [x] Developers

---

## Business & User Impact

* **Impact level:** Critical
* **User reach:** Enterprise customers

The model provides:

* Predictable recurring revenue
* Simple sales conversations
* Easy forecasting
* Automatic scaling with customer portfolios
* Minimal billing complexity

---

## Proposed Solution

Add:

**Account Settings → Enterprise Rate Plan**

### Current Plan

```text
ENTERPRISE

$49 / property / month

Your Portfolio
300 active properties

Monthly Subscription
$14,700
```

---

### Enterprise

**$49 per active property per month**

There are no pricing tiers.

```text
Active Properties × $49
= Monthly Subscription
```

Examples:

```text
50 properties
× $49
= $2,450 / month
```

```text
100 properties
× $49
= $4,900 / month
```

```text
300 properties
× $49
= $14,700 / month
```

```text
1,000 properties
× $49
= $49,000 / month
```

---

## Active Property Definition

Only properties considered **active** for billing are charged.

A property is billable when:

* Property is active
* Property belongs to the enterprise account
* Property is available for management through Domits

Properties marked:

* Draft
* Archived
* Deleted
* Inactive

are not billed.

---

## Account Settings UI

```text
Account Settings
│
└── Enterprise Rate Plan
      │
      ├── Current Plan
[...]

PR #3285, opened 2026-09-21
feat(enterprise): add active property calculator and rate plan ORM mo…

BeMounir:
## Close or Relate the Issue
- Closes #
- Related Issue: #3191

## Proposed Changes
Implemented the core backend logic, database schema, and test suite for the Enterprise Rate Plan feature ($49/active property/month).

Description:
- Created the database migration schema for the `enterprise_rate_plans` table.
- Added the `EnterpriseRatePlan.js` ORM model in `backend/ORM/models/`.
- Implemented `getActivePropertyCount()` in `backend/functions/Alessio-Revenue/activePropertyCalculator.js` to count billable properties (`status: 'active'`, `is_deleted: false`).
- Added unit tests in `test/Alessio-Revenue/activePropertyCalculator.test.js` covering standard property counts, zero active properties, and high-volume property sets using `jest.mock()`.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [ ] Big change (+-max 1000)

## Refactoring
None

## Evidence
*(N/A - Backend logic and unit tests only)*

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop
*(N/A - Non-UI change)*

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated
- [x] Checked for vulnerabilities using "npm audit"

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


elcoleta (changes requested):
The active property calculator looks like it’s using the Property ORM incorrectly. backend/ORM/models/Property.js exports a TypeORM EntitySchema, so Property.count() won’t exist at runtime, and the schema currently doesn’t define enterprise_id or is_deleted either.
The unit test passes because it mocks Property with a fake .count() method, so it doesn’t exercise the real ORM shape. Could we update this to use the TypeORM repository/data source used by the existing Property model, and confirm the enterprise/deleted fields being queried actually exist in the underlying schema?

karimh01 (changes requested):
Agree with Achraf on the ORM and the mock. A few things on top of that, so you have the full list in one go.

**EnterpriseRatePlan is written for Sequelize.** It does require('sequelize') and extends Sequelize's Model, but the repo uses TypeORM and sequelize is not a dependency. And the packages are "type": "module", so require and module.exports fail before any query runs: importing activePropertyCalculator.js gives "require is not defined in ES module scope".

**Status is uppercase.** The property table holds ACTIVE, INACTIVE, ARCHIVED and DRAFT, and every existing check uses "ACTIVE". So status: 'active' would match zero rows even once the query runs.

**The migration will not work on DSQL.** No main. schema prefix, CREATE INDEX without ASYNC, which DSQL rejects, and no .js migration record next to it like the other migrations.

**Import paths.** activePropertyCalculator.js imports ../../ORM/models/..., which does not exist in the deployed zip. The other file already uses database/models/..., which is the right one. It also destructures { EnterpriseRatePlan } from a default export, so that would be undefined.

**The handler has no authentication.** getEnterpriseSubscription.js takes enterpriseId from the path and returns that enterprise's property count and cost to anyone who asks. It is not routed through index.js yet, but it needs auth before it is.

**Money.** price_per_property comes back as a string, and parseFloat plus multiplication gives results like 149.97000000000003, so integer cents would be safer. The default currency is USD while every other money path in the repo is EUR. findOne for the active plan has no ordering, so with two active plans the price is whichever comes first. And nextBillingDate is the fixed string "2026-10-01".

karimh01:
Big improvement. The rate plan model is a proper TypeORM EntitySchema registered in Tables.js, everything is ESM, status is "ACTIVE", the migration has a .js record with main. and CREATE INDEX ASYNC, and the money is integer cents with EUR as default. findOne orders by effective_from and the fixed billing date is gone. The calculator now loads against the real models.

But one new thing is a blocker for the whole platform, not just this feature.

**Property.js now maps enterpriseid and is_deleted, and those columns do not exist.** The migration only creates enterprise_rate_plans, it never adds them to property. TypeORM selects every mapped column on a normal entity read, for example getRepository(Property).find(...) in PriceLabs-Integration/data/repository.js:82. I checked the live table and SELECT property.enterpriseid answers "column does not exist". And because backend/ORM changed, deploy.yml redeploys every Lambda. So the moment this merges, every property read on the platform fails. Either add the columns in a migration that runs before the deploy, or leave the Property entity alone until the enterprise link actually exists.

**The subscription handler cannot load in AWS.** getEnterpriseSubscription.js imports from ../PropertyHandler/auth/... and ../PropertyHandler/data/repository/.... Those are the only cross-function imports in the backend, and the deploy only packages the function's own folder plus .shared, so that path does not exist in the zip. I reproduced the deployed layout and it fails with Cannot find module. Since index.js imports this file at the top, the whole Alessio-Revenue Lambda dies on cold start, including the existing KPI route. Shared code goes through .shared.

**Any host can still read any enterprise's numbers.** authorizeGroupRequest checks the caller is a Host, but enterpriseId still comes from the path with no check that the caller belongs to that enterprise.

**There is still no link between a property and an enterprise.** So even when everything runs, the count can only return zero.

Smaller: effective_until is stored but never used in the plan lookup, so an expired plan still wins as long as its status is active.

karimh01:
This is safe to merge now. Property.js is back to what is on acceptance, so the eight functions that read the property entity generate exactly the SQL they do today. I printed the generated find and insert and ran a real read against the live table through this branch, and neither names the new columns. The only shared change left is the new entity in Tables.js, which is inert with synchronize off. The backend suites that fail on this branch fail the same way on acceptance, so nothing new breaks outside this feature.

is_deleted now has DEFAULT FALSE, and that matters more than before: since the entity no longer maps it, the database default is the only thing that ever sets it. The .sql companion is good too, with the pre-check and the verification query.

Things to know for the follow-up, none of them blocking this PR:

Nothing writes enterpriseid yet. Property creation does not set it, so once the migration runs every row is null, authorizeEnterpriseAccess finds nothing and every request answers 403. Until the migration runs, the count query answers 500 because the column does not exist. Both fail closed, so nothing leaks, but the endpoint cannot return a number to anyone until the property-to-enterprise link exists. Worth saying in the PR description that this is the foundation and the link comes next.

The .js migration still runs two ALTERs, an UPDATE and a CREATE TABLE through one query runner. DSQL rejects that if anything wraps it in a transaction. The .sql companion is the right one to run, statement by statement in autocommit, and it would help to say that in one of the two files.

The API Gateway route already exists on the default stage, but until this deploys, a request to it falls through index.js into the host KPI controller. The development stage was last deployed in November 2025, so the route is not there. And the routing branch in index.js has no test.

elcoleta (approved):
Looks good after the latest fixes. The ORM usage, migration compatibility, authorization, rate-plan selection and money handling are now aligned with the existing backend patterns, and the Property entity is no longer changed in a way that affects existing reads. Approving.

karimh01 (approved):
approved

Issue #3032, opened 2026-05-22
Revenue Management System (RMS): KPI: Gross Missed Revenue


stefanhopman1:
## 🚀 Feature Summary

Introduce a **Gross Missed Revenue KPI** within the Revenue Management System (RMS) to quantify revenue opportunities lost due to suboptimal pricing, unsold inventory, blocked dates, and occupancy gaps.

This KPI helps property managers identify hidden revenue leakage and optimize pricing and availability strategies.

- [ ] v1 Single Listing Formula: Gross Missed Revenue =  Average Daily Rate (ADR) x Unbooked Nights
- [ ] v2 Portfolio Formula: Gross Missed Revenue = Potential Revenue (Optimal Market Rate x Potential Occupied Nights) − Actual Revenue

---

## 🧩 Problem Statement

Most hosts and property managers only track earned revenue, but they lack visibility into how much revenue could have been generated under optimal conditions.

Common causes of missed revenue include:

- Underpriced nights
- Unsold inventory during demand periods
- Last-minute vacancies
- Excessively restrictive minimum stay rules
- Closed arrival/departure restrictions
- Manual pricing errors
- Delayed rate updates

Without measuring missed revenue, significant profit opportunities remain invisible.

---

## ✅ Desired Outcome

Users can understand:

- Total revenue lost during a selected period
- Which properties generated the most missed revenue
- Which pricing and occupancy decisions caused losses
- How much additional revenue could have been captured

---

## 👥 Target Users

- [x] Hosts / Property Managers / Enterprise
- [x] Revenue Managers
- [x] Asset Managers
- [ ] Guests / Customers

---

### Gross Missed Revenue

**Formula:**
Potential Revenue − Actual Revenue = Gross Missed Revenue

Where:
**Potential Revenue**
Optimal Market Rate × Potential Occupied Nights

**Actual Revenue**
Booked Revenue Generated

---

### Portfolio Level

- Gross Missed Revenue (€)
- Gross Missed Revenue (%)
- Revenue Capture Rate (%)

### Property Level

- Missed Revenue by Property
- Missed Revenue by Date
- Missed Revenue by Channel
- Missed Revenue by Rate Plan

---

## 📈 Root Cause Analysis

Identify revenue leakage from:

### Pricing

- Underpriced dates
- Missed dynamic pricing opportunities
- ADR below market benchmarks

### Occupancy

- Unbooked high-demand dates
- Vacancy gaps
- Low occupancy periods

### Restrictions

[...]

PR #3286, opened 2026-09-21
feat(pricelabs): add gross missed revenue KPI endpoint

Nishchal-code:
## Close or Relate the Issue
- Closes #
- Related Issue: #3032

## Proposed Changes
Description:
Backend for a v2-lite version of the Gross Missed Revenue KPI, using PriceLabs' real suggested nightly price instead of the host's flat average ADR (v1, already shipped on the frontend). Scoped down from the full v2 spec: only unbooked nights are counted (no estimation of "underpriced booked nights"), and hosts without an active PriceLabs connection get an explicit `{ connected: false }` rather than a fabricated number.

- Added `MissedRevenueService.getMissedRevenue(hostId, startDate, endDate)`: sums PriceLabs' suggested price over unbooked nights only, per host. Returns `priceDataCoveragePct` alongside the total so partial PriceLabs sync data is never silently presented as complete, plus a per-property breakdown.
- Added `Repository.getCalendarPriceDataForHost`: pulls `pricelabs_price` rows for a host's properties within a date range from the existing `Property_Calendar_Override` table — no schema changes.
- Added `GET /insights/missed-revenue` on the existing PriceLabs-Integration Lambda, defaulting to the current month when no range is given.
- Added `validateDateRange`: rejects malformed dates, reversed ranges, and ranges over 366 days with a 400, before they reach the database (matches the existing validation style in `connect()`).
- 16 new tests: 6 for the missed-revenue calculation (connection gating, unbooked-only filtering, cancelled-booking exclusion, missing-price-data handling, zero-data state, per-property breakdown), 10 for date-range validation.

This went through an eng-review pass before commit: one finding (no input validation on the date range) was raised and fixed in the same commit, along with its test coverage.

Not in scope: frontend wiring to call this endpoint (separate PR), and the full v2 formula (market-rate benchmarking for hosts without PriceLabs, underpriced-booked-night estimation, per-channel/rate-plan breakdowns) — those need real data sources that don't exist yet.

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Small change (less than 350)
- [ ] Big change (+-max 1000)

## Refactoring
None — this is additive only, no existing files' behavior was changed beyond adding new methods/routes.

## Evidence

*(N/A — backend-only, no UI change)*

## Responsiveness
- [ ] UI checked on mobile
- [ ] UI checked on tablet
- [ ] UI checked on desktop

*(N/A — backend-only)*

## Npm Packages
- [ ] NPM Packages installed
- [ ] NPM Packages removed
- [ ] NPM Packages updated

## Checklist
- [ ] Conventions
- [ ] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing

## Keep or delete my branch
- [ ] Keep my branch


karimh01 (changes requested):
The basics are solid. The host id comes from the token and never from the request, so nobody can read another host's numbers. The sum stays in whole integers, so no float drift. And the price is taken per night from property_calendar_override, so a range across a price change is counted correctly. All the models and columns it reads exist on the live schema.

**The number counts nights that are not missed.** The query filters only on host and date range (repository.js:124-134), and the sync writes pricelabs_price for 12 to 18 months ahead. So every unbooked night with a price counts as missed revenue, including:

- nights the host blocked or keeps for an owner stay: is_available = false is never read here, while the v1 path does check it at priceLabsService.js:460
- nights with stop_sell, closed_to_arrival or closed_to_departure
- nights the host told PriceLabs to ignore, pricelabs_ignored = true
- properties that are not live: DRAFT, INACTIVE and ARCHIVED are counted the same as ACTIVE

A host with a blocked month would see that whole month as money lost. The fix is to filter those out in the query, and add a test for each, since none of the tests construct a blocked, stop-sell, ignored or non-ACTIVE row.

**Booking statuses need a decision.** isBookedStatus only treats cancelled, declined and failed as not booked. The live table also has Awaiting Payment, Inquiry, Accepted, Paid and Completed, plus confirmed in three different casings. So right now an Inquiry or an unpaid booking blocks the night from counting as missed. That might be what you want, but it should be a choice, and the tests only use "confirmed" and "cancelled".

**The endpoint is not reachable yet.** The route is in the Lambda, but none of the API Gateway REST APIs has an /insights/missed-revenue resource, so it answers Missing Authentication Token at the edge. It needs the resource, the method and a deployment.

Smaller: the response has no currency field while the v1 endpoints label everything EUR, and the 366-day cap only lives in the controller, so the service itself would scan any range it is given.

karimh01 (approved):
All the main points are fixed. Blocked, stop-sell and ignored nights are now left out, each with its own test, and so are properties that are not ACTIVE. The booked statuses are an allowlist now with the reasoning next to it, which is the safe choice for this KPI, and the test covers every status in the table including the different casings. Currency is in the response and the 366 day limit now also lives in the service.

One correction on my own review: closed_to_arrival and closed_to_departure don't need to be filtered. Those nights can still be booked as part of a longer stay, so leaving them out would undercount missed revenue. Keeping them in is right.

The route doesn't need its own API Gateway resource after all. This API sends everything through the proxy and the Lambda checks the token itself, so it works once this is merged.

Not for this PR: priceLabsService.js:421-424 still uses the old denylist, so an Awaiting Payment night goes to PriceLabs as booked while this KPI counts it as missed. Worth lining those up in a follow-up.

Rakib730066 (approved):
Verified the exclusion logic (blocked/stop-sell/ignored/inactive) and the booked-status allowlist against the test coverage, both match what was discussed.

PR #3287, opened 2026-09-21
Fix/3106 orphan tenant cleanup

karimh01:
## Close or Relate the Issue
- Closes #
- Related Issue: #3106

## Proposed Changes
Description:

When a host deleted a website with a custom domain, the rows went but the CloudFront tenant stayed, enabled and holding the domain. The domain kept showing the Domits unavailable page, and connecting it to a new website failed forever with CNAMEAlreadyExists. The same happened after a lost race, where the loser's disabled tenant kept the hostname.

**On website delete** the site's tenant is disabled, but only one named exactly `dbw-<siteId>` on our distribution. Hand-made tenants like developers-test are never touched. Website deletion only calls the domain service when the row has a tenant id, so environments without the CloudFront variables keep working.

**On reconnect**, when createTenant answers CNAMEAlreadyExists, the tenant holding the domain is looked up. It counts as an orphan only when it is a `dbw-` tenant on our distribution, not named for the site doing the cleanup, and no domain row references its id. Disabled and Deployed: deleted and the create retried once. Still enabled: disabled, and the request answers as before. Anything else is left alone. Recovery for the host is pressing Check again: the verify endpoint provisions any tenant-less row that is not REMOVING, so a second press about fifteen seconds after a disable frees the domain.

**A claim can no longer change the owner of a domain row.** Reviewing this turned up that `ensureDomain` overwrote `site_id` on a conflict, after the ownership check had already run. With the new cleanup that became dangerous: a delayed claim could clear another site's tenant reference, and the cleanup would then disable that site's live tenant. The upsert is now `ON CONFLICT (domain) DO UPDATE ... WHERE existing.site_id = EXCLUDED.site_id` with site_id removed from the SET. A refused claim returns no row and answers domain_taken. This also closes the older takeover race.

**Verified on DSQL** with a rolled-back probe through the same driver the Lambda uses: the WHERE form is accepted, a claim from another site returns zero rows and leaves the row and its tenant reference untouched, and an own-site reclaim returns the updated row.

**Depends on #3259.** The RETURNING fix is merged into this branch because the missing-row guards only work with it. Merge #3259 first. This branch, #3262 and #3270 all touch the ensureDomain statement, so expect a conflict there and resolve it in one place.

**Known, not fixed:** the two fallback-domain callers on publish treat a refused claim as success. That needs a generated hostname to collide with another site's, which requires the same slug and the same eight characters of a random UUID. Before this change such a collision would have silently taken the other site's row, so this is an improvement, but the callers should handle null explicitly in a follow-up.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [ ] Documentation update

## Change Size
- [x] Big change (+-max 1000)

## Refactoring
The failure mapping in createOrAdoptTenant moved unchanged into mapTenantCreateFailure so the retry can reuse it. The diff shows it as removed and re-added.

## Evidence
Live probe on DSQL described above. No UI changes.

## Npm Packages

## Checklist
- [x] Conventions
- [x] Testing
  - [x] Code tested locally
  - [x] Jest tests are included
  - [x] Jest tests are passing, 161 across nine domain suites

## Keep or delete my branch
- [ ] Keep my branch


elcoleta (approved):
Looks good. The orphan cleanup is appropriately constrained to Domits tenants on the configured distribution, existing tenant references are protected, and the conditional domain upsert closes the ownership race without allowing one site to take over another site’s row. The delete/reconnect recovery paths are also covered well. Approving.

BeMounir (approved):

PR #3257, opened 2026-09-16
feat: add Company Public Information section with backend

Mehedihasanb:
## Close or Relate the Issue
- Related Issue: #2798

## Proposed Changes
Description:
Adds the "Public Information" section of the Company settings page (#2798), the first slice of the broader Company epic (Account Status, Public Information, Payouts, Legal Entity). Public Information was chosen as the first slice since it's independent of Finance/Compliance/Payments integration, and the existing stub already had a placeholder for it.

**Frontend:** `CompanyForm.jsx` now has two clearly separated sections — a "Private" section (existing Company name field, used for invoices/reports, not guest-facing) and a new "Public Information" section (Display Name, Logo, Description, Website, Public Email, Public Phone, Country), explicitly labeled "Visible to guests on your listings" so the public/private boundary isn't color-only. Logo upload uses a presigned-URL flow (`useCompanyLogoUpload.js`), following the same pattern already used elsewhere for property images.

**Backend:** new `backend/functions/company-profile/` Lambda (controller/business/data/util layers, following the `communication-preferences`/`host-team` pattern). GET/PUT for the profile, POST for presigned logo-upload URLs. Server-side allow-list validation rejects unknown fields and spoofed `hostId`; required `companyName`; format checks on `website`/`publicEmail`/`publicPhone`. New `Company_Profile` ORM entity and migration.

**Infrastructure (already deployed, not part of this diff but relevant context):** Lambda manually created and deployed in `eu-north-1`, new `/user/company-profile` API Gateway resource with GET/PUT/POST, Cognito-protected, CORS enabled. `main.company_profile` table created in Aurora DSQL. `REACT_APP_COMPANY_PROFILE_API_URL` added to Amplify env vars.

**Known issue, not yet fixed — flagging rather than hiding:** logo upload to S3 succeeds, but the image returns a 403 when the frontend tries to display it. Root cause confirmed: the `accommodation` S3 bucket's public-read policy is scoped to the `images/*` prefix (used by property photos) and does not include the new `company-logos/*` prefix. Fixing this requires either extending the bucket policy (consistent with how property images work) or switching to presigned GET URLs (bigger change, different access model). This affects a shared, already-in-production bucket, so I'm deferring the actual policy change pending sign-off rather than making it unilaterally. Form save, all text fields, and persistence all work correctly — only the logo *display* is affected.

## Change Type
- [x] New feature

## Change Size
[...]

PR #3297, opened 2026-09-22
fix(settings): flaky upload-size tests + missing COMPANY_LOGO_MAX_SIZE constant

Nishchal-code:
## Summary
- `COMPANY_LOGO_MAX_SIZE` was imported in `useCompanyLogoUpload.js` but never exported from `constants.js` — the size guard silently never fired (`file.size > undefined` is always false).
- Both `useCompanyLogoUpload.test.js` and `usePhotoUpload.test.js` faked an oversized file via `Object.defineProperty(file, "size", {value: ...})`, which doesn't reliably stick in GitHub Actions' Linux runner — the override was silently not applying, letting the test fall through into the auto-mocked upload call.
- This single test was failing identically on every open PR's CI (it lives on `acceptance`, not on any individual feature branch), including PR #3261 and others.

## Test plan
- [x] Both fixes verified individually and together locally
- [x] Full frontend suite: 613 passed, 0 failed, 47 skipped (matches CI's pre-existing test count exactly, just without the failure)


Rakib730066 (approved):

EnesVeli (changes requested):
@Nishchal-code thanks for picking this up. One export is still missing: COMPANY_PROFILE_ENDPOINT was dropped by the same merge and is imported in companyProfile.js:2 and companyLogoUpload.js:2. On this branch (2b03cbaf) npm run build fails with Attempted import error: 'COMPANY_PROFILE_ENDPOINT' is not exported from '../constants', and Jest doesn't catch it. Could you add it back as #3257 had it?

export const COMPANY_PROFILE_ENDPOINT = process.env.REACT_APP_COMPANY_PROFILE_API_URL || "/host/company-profile";

On the test change: the logo test failed locally on Windows too, not only in Actions, because COMPANY_LOGO_MAX_SIZE was undefined (file.size > undefined is always false). usePhotoUpload.test.js passed in CI all along. The Uint8Array version is fine to keep, but could the description say the root cause was the missing constant, not the runner? Happy to approve once the build passes.

Mehedihasanb (approved):
approved!

EnesVeli (approved):
thanks, verified.

